PR #29, opened 2015-10-31
expose total_in method, fixes #28

### LINKED ISSUE #28: Reading self-terminating streams
First off, awesome library, thanks!

My use-case: reading git pack files, which contain multiple concatenated, yet undelimited, zlib/deflate streams of data.  In other words, the pack files don't contain any information about the length of the streams - so to detect where the start of the next object in the pack is, I have to know exactly how many bytes the underlying deflate implementation consumed, so I can know where in the underlying file to start reading the next segment of data from.

This poses two separate problems:
* Constructing a ZlibDecoder takes ownership of the underlying file stream, which means I have to re-open the pack file to read the next object from it.  This is an annoyance, but it works fine for my scenario.
* I need to get the number of bytes the ZlibDecoder consumed (which is likely less that the number of bytes it read off the underlying stream, due to buffering).

In an ideal world, I think ZlibDecoder would only take a `&mut` reference to the underlying `Read`, and by some magic, when it's destroyed, it leaves that underlying stream positioned at the exact end of zlib data.  This will likely involve requiring the underling stream to also implement `Seek`, which in turn either requires code duplication in the API (so far as I am aware), or imposes unwanted restrictions on everyone else.  I don't think that's realistic, so let's move on to option 2:

Make ZlibDecoder take a `&mut` reference to the underlying `Read`.  It does nothing special when it's destroyed, but it exposes an extra `zlibDecoder.consumed_bytes()` method (or field, or whatever), calculated from the total bytes it's read, less what's remaining in miniz's input buffer.

The last option for me is to scrap the higher-level API and directly use your miniz-sys bindings, which is ugly for me, but less ugly for everyone else.

Thoughts?

joshuawarner32:
For completeness, it may be good to have the same method on GzDecoder.
I forewent for now since it's not quite as obvious and I don't happen
to need it.  I can take another peek if you think it's important.

### DIFF
--- a/src/deflate.rs
+++ b/src/deflate.rs
@@ -150,6 +150,19 @@ impl<R: Read> DecoderReader<R> {
     pub fn into_inner(self) -> R {
         self.inner.into_inner()
     }
+
+    /// Returns the number of bytes that the decompressor has consumed.
+    ///
+    /// Note that this will likely be smaller than what the decompressor actually
+    /// read from the underlying stream due to buffering.
+    pub fn total_in(&self) -> u64 {
+        self.inner.total_in()
+    }
+
+    /// Returns the number of bytes that the decompressor has produced.
+    pub fn total_out(&self) -> u64 {
+        self.inner.total_out()
+    }
 }
 
 impl<R: Read> Read for DecoderReader<R> {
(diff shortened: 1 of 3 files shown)


src/stream.rs
```diff
@@ -8,7 +8,7 @@ use Compression;
 use ffi;
 
 pub struct Stream<D: Direction> {
-    raw: ffi::mz_stream,
+    pub raw: ffi::mz_stream,
```
alexcrichton: There's a `total_in` method below which perhaps can be used to avoid this `pub`?

alexcrichton:
Looks good to me, thanks @joshuawarner32! Just one small nit but otherwise I'm fine merging

src/stream.rs
joshuawarner32: Ah, good point.  I fix that, renamed consumed_bytes to total_in for consistency, and added total_out for the hell of it.

alexcrichton:
Thanks!

Issue #34, opened 2016-01-04
Could Encoders and Decoders be Send ?


kali:
I don't think miniz stuff would mind being passed around.

alexcrichton:
Hm indeed, I think so!

I'll do an audit and review #36 soon and probably merge

alexcrichton:
Fixed in https://github.com/alexcrichton/flate2-rs/pull/36

PR #36, opened 2016-01-11
make mz_stream Send

kali:
Should answer #34.

### DIFF
--- a/src/stream.rs
+++ b/src/stream.rs
@@ -12,6 +12,9 @@ pub struct Stream<D: Direction> {
     _marker: marker::PhantomData<D>,
 }
 
+unsafe impl<D: Direction> marker::Send for Stream<D> {}
+unsafe impl<D: Direction> marker::Sync for Stream<D> {}
+
 pub enum Compress {}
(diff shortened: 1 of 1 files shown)


alexcrichton:
Ok, I think this is good to go, thanks! 

Can you add the `unsafe impl` to the main crate on `stream::Stream`, however? I believe it should also be safe to declare that as `Sync` as well.

kali:
I'm not sure about Sync really. There are a bunch a pointers and stuff in mz_stream that I would not want to see shared between two threads.

kali:
Here you go. Stream is Send as well.

alexcrichton:
`Sync` itself actually means that `&T` is safe to share among threads, which in this case it is because the type is correctly accessed via `&mut T` and such (e.g. there is no internal mutability). Can you also remove the `Sync` for `mz_stream` directly? (e.g. that doesn't belong in a `*-sys` crate)

kali:
Understood, thanks for the clarifications :)

alexcrichton:
Thanks!

Issue #41, opened 2016-05-02
Premature EOF on WARC files


nostrademons:
When parsing part of the CommonCrawl corpus (which consists of ~1G WARC files where each *record* is individually compressed), flate2 will return EOF after the first chunk has been decompressed rather than continuing to read the rest of the file.  Sample data:

s3://aws-publicdatasets/common-crawl/crawl-data/CC-MAIN-2016-07/segments/1454701166570.91/warc/CC-MAIN-20160205193926-00310-ip-10-236-182-209.ec2.internal.warc.gz

(Downloadable with 'aws s3 cp s3://aws-publicdatasets/common-crawl/crawl-data/CC-MAIN-2016-07/segments/1454701166570.91/warc/CC-MAIN-20160205193926-00310-ip-10-236-182-209.ec2.internal.warc.gz' if you have the AWS CLI installed.  837M, no fees.)

Sample code:

```rust

fn main() {
  let filename = env::args().nth(1).unwrap_or(
      "../CC-MAIN-20160205193926-00310-ip-10-236-182-209.ec2.internal.warc.gz"
      .to_string());
  let input_file = File::open(&filename).unwrap();
  let mut gz_decoder = GzDecoder::new(input_file).unwrap();
  loop {
    let mut buffer = Vec::with_capacity(1000000);
    let capacity = buffer.capacity();
    let read_range = buffer.len() .. capacity;
    unsafe { buffer.set_len(capacity) };
    match gz_decoder.read(&mut buffer[read_range]) {
        Ok(0) => {
          println!("EOF");
          break;
        },
        Ok(_) => {
          println!("Read {}", String::from_utf8_lossy(&buffer))
        },
        Err(err) => {
          println!("Error filling buffer: {}", err)
        }
    }
  }
}
```

Arguably the files shouldn't do this, but the [WARC spec](http://bibnum.bnf.fr/WARC/WARC_ISO_28500_version1_latestdraft.pdf) recommends record-at-a-time compression, and it's pretty common practice in the Hadoop world to operate on big files that are the concatenation of individually-gzipped records so that Hadoop can split the input without reading it.  gunzip/gzcat can read it, and re-compressing it with gzip allows flate2 to as well.  Given that these files exist, maybe flate2 could avoid returning EOF until the underlying stream does, instead returning the stream of decompressed bytes from the next record?

alexcrichton:
[...]

PR #43, opened 2016-06-14
Implemented multi member gz decoder

veldsla:
You mentioned in issue #41 that you were considering creating a separate type for the multistream gzip files. The nested loop method I used to parse these files in https://github.com/alexcrichton/flate2-rs/issues/41#issuecomment-219058833 becomes a bit convoluted when parsing multiple files.

Considering my rust beginner level I used it as an exercise. It works by calling finish_member instead of finish and checking if there are more bytes remaining in the buffer. If this is the case a new header read is attempted and the decoder is reset for the next member. Would this be what you had in mind?


alexcrichton:
@veldsla I'm so sorry this sat here for so long! I had no idea this was here :(

In any case this looks excellent, thanks so much for implementing this! In the future feel free to ping me on any PR that sits for awhile!

veldsla:
Haha, that's great. You did reference this PR in #44, so I just assumed you didn't want to add more functionality. Oh well assumptions are...😄

alexcrichton:
Weird... I maybe accidentally archived the email?

pmarks:
@alexcrichton - any chance we could get a crates.io release w/ this functionality?  Would like to start encouraging downstream packages to use MultiGzDecoder.  Many uses cases where MultiGz files are the norm (especially bioinformatics).

veldsla:
There has been a library evaluation which among other things raised a point about the documentation (#87). I also noticed a doc has been swapped between a multi/normal reader. I will submit a PR for this. I'm guessing the next release will come when #89 is done?

alexcrichton:
@pmarks certainly! I'll do a release right now

PR #52, opened 2016-11-03
Remove unnecessary trait bounds from reader types definitions

l1048576:
It is not necessary for reader types to bind type parameters with `std::io::Read` at the type definition (for example, see [std::io::Take](https://doc.rust-lang.org/std/io/struct.Take.html)).

Writer types requires restrictions such as `W: Write` at the type definition, but reader types do not.
These needless restrictions should be removed.


alexcrichton:
Thanks!

PR #56, opened 2017-01-14
Fix overflows with large inputs on 64-bit Windows

kornholi:
On Windows the counters provided by miniz stay 32-bit wide and cause issues with 4GB+ files.

Not sure of the best approach here. Thoughts?


alexcrichton:
Ah well kind of a bummer we have to do this, but alas! Looks good to me, thanks for the PR!

PR #60, opened 2017-01-23
Expose the underlying mz_stream from Compress and Decompress.

luser:
zlib exposes more API surface than miniz, and I needed to use a few of those APIs (like inflateReset),
but using the libz-sys crate directly means I have to reimplement a number of things from flate2.

Really all I need is to be able to get the underlying mz_stream so I can call the zlib APIs on it.
This patch adds a `get_raw` method to Compress and Decompress that returns a `*mut mz_state` which can
be passed directly to the functions defined in libz-sys. This doesn't add much complexity to flate2, but
fixes my niche use case.

### DIFF
--- a/src/lib.rs
+++ b/src/lib.rs
@@ -83,6 +83,11 @@ pub mod bufread {
     pub use gz::DecoderReaderBuf as GzDecoder;
 }
 
+/// Raw C structs from miniz or libz.
+pub mod raw {
+    pub use ffi::mz_stream;
+}
+
 fn _assert_send_sync() {
     fn _assert_send_sync<T: Send + Sync>() {}
 
--- a/src/mem.rs
+++ b/src/mem.rs
@@ -170,6 +170,11 @@ impl Compress {
         }
     }
 
+    /// Get the raw state of the underlying compressor.
+    pub fn get_raw(&mut self) -> *mut ffi::mz_stream {
+        &mut self.inner.raw as *mut ffi::mz_stream
+    }
+
     /// Returns the total number of input bytes which have been processed by
     /// this compression object.
     pub fn total_in(&self) -> u64 {
@@ -280,6 +285,11 @@ impl Decompress {
                 },
             }
         }
+    }
+
+    /// Get the raw state of the underlying decompressor.
+    pub fn get_raw(&mut self) -> *mut ffi::mz_stream {
+        &mut self.inner.raw as *mut ffi::mz_stream
     }
 
     /// Returns the total number of input bytes which have been processed by


alexcrichton:
Neat! I'm all for exposing features, but my worry here is that exporting the underlying FFI type can cause subtle breakage, especially with the features of this crate. Right now there's a Cargo feature which enables zlib instead of miniz, but in general Cargo features must be additive in terms of APIs in order to be composable. In this case though it'd be changing an API b/c the underlying stream would be changing.

Perhaps helper methods could be added depending on which backend is in use? Or just make the helper methods a noop on one backend but not the other? I'd be totally down with implementing something like `reset` as an inherent method here as well.

luser:
I originally started adding a few helper methods that were `#[cfg(feature = "zlib")]` but then I wanted to use `deflateResetKeep` which isn't even exported in the libz-sys API (but the symbol still exists) and I wasn't sure I wanted to fall down that rabbit hole. :)

I'd be fine with this being something like:
```rust
#[cfg(feature = "zlib")]
pub fn as_raw_zlib(&mut self) -> Option<*mut ffi::mz_stream> {
  Some(&mut self.inner.raw as *mut ffi::mz_stream)
}

#[cfg(not(feature = "zlib"))]
pub fn as_raw_zlib(&mut self)  -> Option<*mut ffi::mz_stream> { None }
```

Does that address your concerns?

alexcrichton:
Yeah that'd solve my concerns about the types for sure (and is also a neat idea!)

I started the libz-sys bindings and such very conservatively, so I'd be totally down for adding all number of bells and whistles to them if you'd like. If you know the APIs you want I could also take care of that!

alexcrichton:
Hm so actually it may not actually solve my concern because while crates would compile they semantically wouldn't work together still. One crate expecting zlib and another expecting miniz would be fundamentally not composable, so at runtime one would likely panic...

If it's just a matter of adding a few apis to libz-sys and then binding them here though that shouldn't be too bad!

luser:
Is it possible to use two crates together if they both use flate2 but one of them specifies `features = ["zlib"]`? I thought cargo would not allow that. (That's a separately-unfortunate state of affairs.)  In this case my crate is toggling that feature for flate2, so my crate should be able to rely on flate2 using zlib.

More specifically, the code I'm using this for depends on these zlib features and won't work without them. If there's no way to leverage the extra features of zlib in flate2 then it seems like having it as an option is not very useful.

Like I said in my initial comment,  I can use `libz-sys` directly for this, but that means I have to rewrite all the interface code you've written in flate2 to wrap the zlib API, which isn't appealing.

I'm still struggling to understand your composability issue--you're worried that two crates that both use flate2, where one might pass a flate2 struct to another, would not work if one expected zlib and one expected miniz? I don't see how that'd work even with the existing code (even if cargo lets you build a project like that, per above)--one crate would be calling zlib functions on the structs, and one would be calling miniz functions on them. Is there some other potential situation here that I'm not getting?

alexcrichton:
Ah yeah so the way Cargo works is that it's got a bunch of dependency edges from a crate to flate2. Each edge can enable some number of features, by default all edges enable the "default" feature. The flate2 crate is the compiled **once** with a union of all sets of enabled features. This means that features (in the Cargo sense) can only be *additive* to compose well together. Features currently can't be mutually exclusive or anything like that.

So in that sense if you enable the zlib feature, you'll always get an implementation backed by zlib (no need to worry about that!). Crates relying on miniz would be broken, but that seems like a far-fetched case.

Now if we were to expose these raw functions, imagine that zlib has special functionality but miniz also has special functionality. One crate uses the zlib functionality and one crate use miniz special functionality. At that point those two crates fundamentally can't be linked together through the flate2 crate.

My preference here would be to bind everything in the flate2 crate itself. If the APIs can't be implemented in miniz that's fine, zlib can still be an additive feature that enables the API where miniz just doesn't have an impl yet. Does that make sense? Earlier I [added a binding for `inflateReset2`](https://github.com/alexcrichton/flate2-rs/commit/93c9806be3eaf58a925f2905196a5ec77495bfff) as an example.

luser:
OK, so something like the following would be OK?

```rust
#[cfg(feature = "zlib")]
pub fn reset_keep(&mut self) {
 // ...
}
```

Where this method just doesn't exist in builds without the zlib feature?

alexcrichton:
Yeah that sounds reasonable to me!

alexcrichton:
Ok I'm going to close this due to the brekage aspect, but I'm more than willing to add features with `zlib`!

(e.g. add methods behind the `zlib` feature)

PR #63, opened 2017-03-04
Replace libc with std::os::raw

serprex:


alexcrichton:
Thanks for the PR! Could you expand on the motivation here? As a crate compiling C code, this crate *does* depend on libc?

serprex:
[Spurred by discussion](https://www.reddit.com/r/rust/comments/5xb0d5/rust_roadmap_libc_v10_release/dehfexq), this makes flate2-rs agnostic towards libc version in use

This crate may depend on C's libc, but it needn't depend on Rust's libc

alexcrichton:
Ok, in that case I'm going to close this as the libc crate is *the* way to link to a C library on a given platform, which is what this is embodying. Thanks for the PR though!

nox:
This doesn't mean libc must be a public dependency, @alexcrichton would you be ok with keeping the dependency but not relying on its symbols from the public API? This will limit the damage of a libc 1.0 bump.

PR #65, opened 2017-03-20
Add flush_finish()

vandenoever:
This function is useful for creating deflated blocks that can be concatenated.

And that in turn is useful for deflating a stream in parallel.

### DIFF
--- a/src/deflate.rs
+++ b/src/deflate.rs
@@ -89,10 +89,22 @@ impl<W: Write> EncoderWriter<W> {
 
     /// Consumes this encoder, flushing the output stream.
     ///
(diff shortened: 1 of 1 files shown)


alexcrichton:
Hm I think I may be missing something, this looks like it has the same implementation of `into_inner` above it?

vandenoever:
It uses `flush()` instead of `finish()`.
By using `flush()`, it's possible to add more later. Or combine bytearrays created in different threads.

alexcrichton:
Heh I think my brain saw `f***sh` in both cases and filled in the blank to say "surely this is the same!"

In any case looks great, thanks!

Issue #62, opened 2017-02-22
Implement get_ref/get_mut/total_in/total_out/into_inner for gzip DecoderReader


sid0:
The DEFLATE `DecoderReader` has a bunch of useful methods. It'd be nice to have the same methods on the gzip `DecoderReader` as well.

Relatedly, it would also be really nice to have traits that all the decoders and encoders here (and in your `bzip2-rs` repo, but not sure how easy that would be) implement, so that users could (statically or dynamically) switch out compression engines :)

alexcrichton:
Sounds good to me! I may not have time to implement this soon but it should be a relatively easy PR!

sid0:
Any thoughts on exposing a common interface? (Right now I'm doing it within my own project.)

Still working through some internal paperwork required to contribute :)

alexcrichton:
I'd personally prefer to hold off on that for now, but if it ends up being popular we can certainly add it!

Let me know if you need any help though!

alexcrichton:
I believe this was effectively done in #66, so closing

PR #66, opened 2017-03-24
Implement get_ref, get_mut and into_inner for gzip encoders and decoders

sid0:
Partly addresses #62.


alexcrichton:
Looks great!

PR #69, opened 2017-03-29
zlib requires the stream to have a stable memory location.

Lukazoid:
As the zlib library internal state stores a pointer back to the stream (see [deflate.h](https://github.com/alexcrichton/libz-sys/blob/master/src/zlib-1.2.8/deflate.h#L98)), this requires the stream to have a stable memory location and never change.

This commit introduces a stream wrapper which heap allocates when using zlib but not when using miniz.

It seems like a lot of zlib's methods do not rely upon this pointer back however some do, I first observed this issue when attempting to add an implementation of `deflateSetDictionary`.

### DIFF
--- a/src/ffi.rs
+++ b/src/ffi.rs
@@ -5,6 +5,7 @@ pub use self::imp::*;
 mod imp {
     extern crate libz_sys as z;
     use std::mem;
+    use std::ops::{Deref, DerefMut};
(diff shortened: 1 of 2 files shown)


alexcrichton:
Oh when I first added these bindings I was wondering if that was a requirement, turns out it is!

Thanks for the PR, looks great!

PR #70, opened 2017-03-29
Added a .gitattributes to configure how files are handled in git.

Lukazoid:
Checking out flate2-rs under windows with autocrlf=true results in some of the integration tests failing. This is because the uncompressed text files are undergoing end-of-line normalization while the archive contents still contain the LF line-endings, this results in the test assertions failing.

By using a .gitattributes file it is possible to ensure the integration test text files do not undergo end-of-line normalization.

### DIFF
--- a/.gitattributes
+++ b/.gitattributes
@@ -1,0 +1,20 @@
+# automatically handle text files
+* text=auto
+
+# text files
+*.md text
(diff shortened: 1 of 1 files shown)


alexcrichton:
Thanks!

PR #74, opened 2017-04-03
Exposed deflateSetDictionary and inflateSetDictionary functionality when using zlib

Lukazoid:
The functionality is currently only available when using the `Compress` and `Decompress` types, deciding upon a usable API for the IO types requires a fair bit more thought and may be something to consider in future.


alexcrichton:
Thanks for the PR!

I'm somewhat wary about making this zlib only, but it seems fundamentally incompatible (or unimplemented) with miniz, so in that sense seems fine!

Would it be possible to avoid extending the error enum for now? That way we could land this immediately, otherwise it'll need to wait until 1.0

RReverser:
I'd love to have access to dictionary API too. @alexcrichton is your question still the primary blocker of this PR? @Lukazoid do you mind either updating it or may I use your branch as base for a new PR?

alexcrichton:
@RReverser ah yeah now that this crate has hit 1.0 it'd need to be added in a backwards-compatible fashion, but I'd still be ok landing!

oyvindln:
It should be relatively simple to add support for this in the rust back-end later as well.

RReverser:
@alexcrichton To clarify: by "backwards-compatible fashion" you mean that extending `Status` enum is considered non-backward-compatible, is that correct?

If so, do you mind if I make this part of `DecompressError` - it's opaque so I guess changing internal `()` to enum and exposing something like `err.needs_dictionary() -> Option<u32>` should be fine?

alexcrichton:
@RReverser yeah, but exposing it as an error sounds great!

Lukazoid:
Sorry I never got back to you when I originally made this PR @alexcrichton. If @RReverser hasn't implemented this functionality already, I can make those changes (add the `needs_dictionary` to `DecompressError`) later today.

RReverser:
@Lukazoid I haven't gotten to it either (beyond some local changes to bring it up-to-date with master), so feel free to update your PR - that's even better.

src/mem.rs
```diff
@@ -218,6 +234,21 @@ impl Compress {
...
+    ///
+    /// Returns the Adler-32 checksum of the dictionary.
+    #[cfg(feature = "zlib")]
+    pub fn set_dictionary(&mut self, dictionary: &[u8]) -> u32 {
+        let stream = &mut *self.inner.stream_wrapper;
+        let rc = unsafe {
+            ffi::deflateSetDictionary(stream, dictionary.as_ptr(), dictionary.len() as ffi::uInt)
+        };
+
+        assert_eq!(rc, ffi::MZ_OK);
```
RReverser: I was thinking it would be better to return `Result<u32, DecompressError>` from this function so that error in case of incorrect dictionary or in case of calling `set_dictionary` at incorrect time (it has to be called only immediately after `Z_NEED_DICT`) could be still handled gracefully.

src/mem.rs
```diff
@@ -419,6 +453,17 @@ impl Decompress {
...
 
+    /// Specifies the decompression dictionary to use.
+    #[cfg(feature = "zlib")]
+    pub fn set_dictionary(&mut self, dictionary: &[u8]) {
+        let stream = &mut *self.inner.stream_wrapper;
+        let rc = unsafe {
+            ffi::inflateSetDictionary(stream, dictionary.as_ptr(), dictionary.len() as ffi::uInt)
+        };
+
+        assert_eq!(rc, ffi::MZ_OK);
```
RReverser: I was thinking it would be better to return `Result<u32, DecompressError>` from this function so that error in case of incorrect dictionary or in case of calling `set_dictionary` at incorrect time (it has to be called only immediately after `Z_NEED_DICT`) could be still handled gracefully.

src/mem.rs
RReverser: Maybe return `Result` at least as `Result<u32, ()>`?

src/mem.rs
Lukazoid: I believe I originally implemented these methods before flate2 returned `Result` and simply had asserts against MZ_OK.

Returning an actual `Result` is definitely better, I assume it's safe to presume the adler code will already be populated from a previous call to inflate. So if `inflateSetDictionary` works without error I can return that.

alexcrichton:
This looks awesome, thanks so much @Lukazoid!

Issue #75, opened 2017-04-04
Rename internal types to match the public types


dtolnay:
The write/read/bufread modules all repeat the same type names:

```rust    
pub use gz::EncoderReader as flate2::read::GzEncoder;
pub use deflate::EncoderReader as flate2::read::DeflateEncoder;
```

Unfortunately the [rustdoc for `flate2::read::GzEncoder`](https://docs.rs/flate2/0.2.17/flate2/read/struct.GzEncoder.html) shows the private names:

```rust
impl<R: Read> EncoderReader<R>
fn new(r: R, level: Compression) -> EncoderReader<R>
```

This is a rustdoc bug but even aside from that, this pattern is confusing to people browsing the code because what they see clicking through a [src] link does not match how they will be using the library. Let's try to make the real type names match the public names they are exported as.

Relevant API guideline: https://github.com/brson/rust-api-guidelines/issues/5
Rustdoc bug: https://github.com/rust-lang/rust/issues/41072

opilar:
I think it would be neat to have the same name for the same functionality but for different modules:
```rust
read::deflate::Encoder
read::zlib::Encoder
read::gz::Encoder
```
So the module declaration would be: 
```rust 
pub mod read {
    pub use deflate;
    pub use zlib;
    pub use gz;
}
```

alexcrichton:
@opilar yeah that seems plausible, but other compression/decompression formats probably wouldn't want to mirror that (e.g. `read::xz::Decoder`). Basically for consistency across crates (e.g. bzip2, xz2, brotli2, etc) I'd favor dropping the inner `deflate`, `zlib`, and `gz` modules here

opilar:
@alexcrichton so all encoders and decoders goes to one `lib.rs` file? That's not good.

alexcrichton:
Oh no I think they'll be in `src/read.rs`, `src/write.rs`, and `src/bufread.rs`

alexcrichton:
(oops didn't mean to close)

sw12:
Hi! I'm interested in working on this (it'll be my first Rust OSS contribution). After reading through the code, I have a couple questions. I understand the goal to be to restructure the modules `{deflact,zlib,gz}.rs` -> `{read,bufread,write}.rs`. . If that's correct, where would you like the `Builder` and `Header` structs, and the `test` submodule, within each of the existing modules to live?

I poked around your other compression crates to try and find a convention but I found none. I was thinking perhaps either

[...]

PR #94, opened 2017-04-11
Extend MultiGzDecoder docs

### LINKED ISSUE #87: Explain what a gzip multistream means
All the documentation currently says is:

> A gzip streaming decoder that decodes all members of a multistream.

As someone who wants to decode some gz data but is not familiar with gz spec details, I can't tell from this whether MultiGzDecoder is relevant to my use case.

In contrast, the Go gzip package documentation is much more helpful:

> The Reader expects the input to be a sequence of individually gzipped data streams, each with its own header and trailer, ending at EOF. The effect is that the concatenation of a sequence of gzipped files is treated as equivalent to the gzip of the concatenation of the sequence. This is standard behavior for gzip readers.

veldsla:
This PR adds an explanation to the MultiGzDecoder. Love the wording in the evaluation :smile:
> I didn't know what a gzip multistream is so I searched for "gzip multistream" and landed on the Go gzip package's documentation which explained everything I needed to know. Then I felt dirty.

In my defense :wink: I picked the multistream term _because_ it was in the Go docs, my personal preference was `MultiMember`. We can of course still change this!

I'm a bit uncertain if I should refer to the public name `MultiGzDecoder`, or the internal name `MultiDecoderReader`, but this is also related to #75.

This fixes #87.

### DIFF
--- a/src/gz.rs
+++ b/src/gz.rs
@@ -71,16 +71,23 @@ pub struct DecoderReader<R: Read> {
 
 /// A gzip streaming decoder that decodes all members of a multistream
 ///
-/// This structure exposes a `Read` interface that will consume compressed
-/// data from the underlying reader and emit uncompressed data.
+/// A gzip member consists of a header, compressed data and a trailer. The [gzip
+/// specification](https://tools.ietf.org/html/rfc1952), however, allows multiple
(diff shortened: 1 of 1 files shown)


alexcrichton:
Awesome, thanks @veldsla!

PR #95, opened 2017-04-12
Use Unix style (LF) line endings

SamWhited:
On a Unix like system if you clone this repo the line endings in miniz.c are broken by the .gitattributes file; you immediately have pending changes. Swapping the line endings to be native should fix this (I think Windows machines will rewrite the line endings in the working directory, and then go back to Unix style on commit).

This does break git-blame quiet badly, which is annoying, but probably less annoying than always having pending changes in your working directory that you could accidentally add to any random commit.

(it may be that this can be fixed in gitattributes actually to work on Windows and Unix; I generally tell people to use Unix style line endings and let Windows people configure Git locally instead of forcing a config on everyone who clones your repo, but I don't know enough about it to be sure if this is viable, and don't have a Windows machine to test with anyways)


alexcrichton:
Thanks!

Issue #84, opened 2017-04-04
GzBuilder methods should take Into<Vec<u8>>


dtolnay:
Currently [`GzBuilder::filename`](https://docs.rs/flate2/0.2.17/flate2/struct.GzBuilder.html#method.filename) and [`GzBuilder::comment`](https://docs.rs/flate2/0.2.17/flate2/struct.GzBuilder.html#method.comment) take `&[u8]` while [`GzBuilder::extra`](https://docs.rs/flate2/0.2.17/flate2/struct.GzBuilder.html#method.extra) takes `Vec<u8>`.

These would be more consistent and more usable if they all accepted `Into<Vec<u8>>`.

alexcrichton:
Fixed in #99

PR #96, opened 2017-04-13
GzBuilder methods now take Into<Vec<u8>>.

Matt8898:
Solves issue #84.

### DIFF
--- a/src/gz.rs
+++ b/src/gz.rs
@@ -146,8 +146,8 @@ impl Builder {
     }
 
     /// Configure the `extra` field in the gzip header.
-    pub fn extra(mut self, extra: Vec<u8>) -> Builder {
-        self.extra = Some(extra);
+    pub fn extra<T: Into<Vec<u8>>>(mut self, extra: T) -> Builder {
+        self.extra = Some(extra.into());
         self
     }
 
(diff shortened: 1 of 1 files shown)


src/gz.rs
```diff
@@ -998,10 +998,10 @@ mod tests {
 
     #[test]
     fn fields() {
-        let r = vec![0, 2, 4, 6];
+		let r = vec![0, 2, 4, 6];
         let e = Builder::new()
-                    .filename(b"foo.rs")
-                    .comment(b"bar")
+                    .filename(&b"foo.rs"[..])
+                    .comment(&b"bar"[..])
```
alexcrichton: Hm I think that this works if the `b` is removed here, right?

src/gz.rs
```diff
@@ -998,10 +998,10 @@ mod tests {
 
     #[test]
     fn fields() {
-        let r = vec![0, 2, 4, 6];
+		let r = vec![0, 2, 4, 6];
```
alexcrichton: Looks like some extra tabs may have snuck in?

alexcrichton:
Thanks @Matt8898!

Matt8898:
Yes, that works without b.

Matt8898:
I'm going to change that as soon as possible.

PR #97, opened 2017-04-13
GzHeader::mtime documentation clarification

### LINKED ISSUE #85: Clarify documentation of GzHeader::mtime
> Returns the \`mtime\` field of this gzip stream's header, if present.

The mtime field is always required to be present. But it is 32 bits so because of the 2038 problem its use is discouraged. Most libraries set it to 0 and never look at it.

Document that the usage of mtime is discouraged and that 0 is the expected value for modern libraries.

opilar:
`mtime` info was taken from RFC 1952:
http://www.zlib.org/rfc-gzip.html#member-format

Resolves #85.

### DIFF
--- a/src/gz.rs
+++ b/src/gz.rs
@@ -841,7 +841,15 @@ impl Header {
         self.comment.as_ref().map(|s| &s[..])
     }
 
-    /// Returns the `mtime` field of this gzip stream's header, if present.
+    /// This gives the most recent modification time of the original file being compressed.
+    ///
+    /// The time is in Unix format, i.e., seconds since 00:00:00 GMT, Jan. 1, 1970.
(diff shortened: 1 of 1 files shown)


alexcrichton:
Thanks for the PR! Maybe we should hold off on `#[deprecated]` though? It seems like modern usage of that field would mostly be forced by some other implementation (or legacy system) in which case the deprecation warning may not help much :(

opilar:
@alexcrichton Sure. Should I just documented that this is discourage usage?

Also I've found that `mtime` should be signed integer.

alexcrichton:
Yeah the documentation here looks good to me (discouraging its use).

Let's hold off on the changes to the sign though, I'd prefer to avoid breaking changes just yet. (we can batch up later)

alexcrichton:
Thanks!

PR #98, opened 2017-04-13
Document the GzBuilder panic cases

### LINKED ISSUE #77: Document the GzBuilder panic cases
[`GzBuilder::filename`](https://docs.rs/flate2/0.2/flate2/struct.GzBuilder.html#method.filename) and [`GzBuilder::comment`](https://docs.rs/flate2/0.2/flate2/struct.GzBuilder.html#method.comment) both panic if the bytes contain 0.

Document these panics in a "Panics" section of the doc comment of those methods.

opilar:
Resolves #77.

### DIFF
--- a/src/gz.rs
+++ b/src/gz.rs
@@ -152,12 +152,20 @@ impl Builder {
     }
 
     /// Configure the `filename` field in the gzip header.
+    ///
+    /// # Panics
+    ///
+    /// Panics if `filename` argument is empty slice.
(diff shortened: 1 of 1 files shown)


alexcrichton:
Thanks! I think the panic condition here is "if the slice contains a zero" rather than an empty slice, though

opilar:
@alexcrichton Looks good now?

alexcrichton:
Thanks!

PR #103, opened 2017-04-17
add a method to get mtime of a GzHeader as a datetime

### LINKED ISSUE #86: Method to get mtime of a GzHeader as a datetime
There is no meaning formally associated with mtime but number of seconds since the epoch seems like a common interpretation. Add a method to get the mtime of a GzHeader as some datetime type assuming the value represents seconds since the epoch.

nivkner:
fixes #86

### DIFF
--- a/src/gz.rs
+++ b/src/gz.rs
@@ -8,6 +8,7 @@ use std::ffi::CString;
 use std::io::prelude::*;
 use std::io;
 use std::mem;
+use std::time;
 
 #[cfg(feature = "tokio")]
 use futures::Poll;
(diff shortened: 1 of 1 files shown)


src/gz.rs
```diff
@@ -862,6 +863,17 @@ impl Header {
     pub fn mtime(&self) -> u32 {
         self.mtime
     }
+
+    /// Returns the most recent modification time as a datetime if time stamp is available.
+    pub fn datetime(&self) -> Option<time::SystemTime> {
```
alexcrichton: Perhaps this could be called something like `mtime_as_datetime` to clarify that this is just reading the `mtime` field of the header?

src/gz.rs
```diff
@@ -862,6 +863,17 @@ impl Header {
     pub fn mtime(&self) -> u32 {
         self.mtime
     }
+
+    /// Returns the most recent modification time as a datetime if time stamp is available.
```
alexcrichton: Could you expand the documentation here as well to explain when `None` is returned and to detail how the `mtime` field is interpreted?

alexcrichton:
Looks great to me, thanks @nivkner!

src/gz.rs
nivkner: the interpretation of `mtime` is already in the associated method,  so do I also duplicate the parts about year 2038 and local vs universal,  or just the meaning of it as time since the epoch?

nivkner:
I fixed my PR according to @alexcrichton's review. I added the explanation needed to use the method: explain the case where there is `None` and what the time stands for. However I didn't want to copy&paste an entire paragraph on the same page, so I just linked to it (which simply highlights it right now).

Also I changed the name. Would have done it in the first place, but I just wasn't sure about the conventions surrounding alternate representations of struct fields.

alexcrichton:
Perfect, thanks @nivkner!

PR #104, opened 2017-04-17
Remove  bound on structs

### LINKED ISSUE #88: Remove R: Read bounds on structs
None of the `R: Read` bounds are required. This includes the following types:

- deflate::EncoderReader
- deflate::EncoderReaderBuf
- deflate::DecoderReader
- deflate::DecoderReaderBuf
- gz::EncoderReader
- gz::EncoderReaderBuf
- gz::DecoderReader
- gz::MultiDecoderReader
- gz::DecoderReaderBuf
- gz::MultiDecoderReaderBuf

All of these need `R: Read` bounds on various impls but not on the structs themselves. Having bounds on the structs is annoying because the bounds transitively infect any structs containing any of these types.

The `W: Write` bounds in contrast are all required because those types have Drop impls that require the bound. Rust currently requires any bound on a Drop implementation to also be present on the struct.

opilar:
Resolves #88

### DIFF
--- a/src/bufreader.rs
+++ b/src/bufreader.rs
@@ -20,7 +20,7 @@ pub struct BufReader<R> {
     cap: usize,
 }
 
-impl<R: Read> BufReader<R> {
+impl<R> BufReader<R> {
     pub fn new(inner: R) -> BufReader<R> {
         BufReader::with_buf(vec![0; 32 * 1024], inner)
     }
--- a/src/crc.rs
+++ b/src/crc.rs
@@ -60,7 +60,7 @@ impl Crc {
     }
 }
 
-impl<R: Read> CrcReader<R> {
+impl<R> CrcReader<R> {
     /// Create a new CrcReader.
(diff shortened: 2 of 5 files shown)


alexcrichton:
Thanks for the PR! While you're at it would you mind updating a few methods as well?

Right now the constructors should require bounds like `R: Read`, but methods like `get_ref`, `into_inner`, etc ideally wouldn't Could you move those methods to a separate `impl` block without the bounds of `R: Read`?

If you're busy though please just let me know! I certainly don't mind merging this ahead of that :)

opilar:
@alexcrichton I hope I didn't miss anything. Where can I find some resources about using bounds properly. I've read articles about traits but nothing about this issue.

alexcrichton:
Thanks for the PR! Looks great.

Right now this is mostly just a conventions issue in terms of where the bounds are located, but we shoudl have the conventions written up soon!

Issue #83, opened 2017-04-04
All public types should implement Debug


dtolnay:

alexcrichton:
Fixed in https://github.com/alexcrichton/flate2-rs/pull/105

PR #105, opened 2017-04-18
Implement Debug for public types #83

kper:
Is the message for the `impl Debug` in `src/ffi.rs` line 111 okay?

### DIFF
--- a/src/bufreader.rs
+++ b/src/bufreader.rs
@@ -13,12 +13,20 @@ use std::io;
 use std::io::prelude::*;
 use std::mem;
 
-#[derive(Debug)]
 pub struct BufReader<R> {
     inner: R,
     buf: Box<[u8]>,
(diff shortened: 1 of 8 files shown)


src/bufreader.rs
```diff
@@ -13,6 +13,7 @@ use std::io;
 use std::io::prelude::*;
 use std::mem;
 
+#[derive(Debug)]
```
alexcrichton: Could this type actually have a custom debug impl like `BufReader` in the standard library? That way we can avoid accidentally dumping out kilobytes of raw data :)

alexcrichton:
Everything looks great to me, thanks so much @kper! Just one small nit but otherwise looks ready to merge.

alexcrichton:
Thanks again for the PR!

Issue #79, opened 2017-04-04
Use distinct Flush types for Compress::compress vs Decompress::decompress


dtolnay:
Currently there is a single [`flate2::Flush`](https://docs.rs/flate2/0.2.17/flate2/enum.Flush.html) type that is used by both [`flate2::Compress::compress`](https://docs.rs/flate2/0.2.17/flate2/struct.Compress.html#method.compress) and [`flate2::Decompress::decompress`](https://docs.rs/flate2/0.2.17/flate2/struct.Decompress.html#method.decompress).

The methods are different in that `compress` supports Flush::None, Sync, Partial, Full, Finish while `decompress` supports Flush::None, Sync, Finish. Neither method supports Flush::Block.

It would be nicer to enforce these restrictions statically by having two different Flush types with only the variants that are allowed in each case.

SamWhited:
What would happen to the `Ops` trait in this situation? If you give it an associated type `Flush`, I don't think you'll be able to use any of the enum values in the `Writer` implementation. Would Writer be split out into a `CompressWriter` and `DecompressWriter`? Trying to make `Ops`/`Writer` generic  WRT a flush type gets very complicated very quickly; I'm not sure there's a good way to do this without a lot of duplication.

alexcrichton:
Eh I'd be fine with a bit of duplication in that respect @SamWhited, that's what we have tests for!

chrisvittal:
I'm willing to take care of this. Here's my first pass at a design for this.

1. Split `Flush` into `FlushCompress` and `FlushDecompress` with
    + `pub enum FlushCompress { None, Sync, Partial, Full, Finish }`
    + `pub enum FlushDecompress { None, Sync, Finish }`
2. Add a private `Flush` trait that is used to construct `FlushCompress` or `FlushDecompress` values.
    + The `Flush` trait has methods `none()`, `sync()`, `partial()`, `full()`, and `finish()`, all returning `Self`
    + The `impl Flush for FlushDecompress` panics for `partial()` and `full()` as they are not valid for `FlushDecompress`
3. Add the associated type `Flush: Flush` to `Ops` with the impls for `Compress` having `Flush = FlushCompress` and `Decompress` having `Flush = FlushDecompress` respectively.
4. Change all references to `Flush::MEMBER` to the appropriate `Flush` constructor method.
    + For example change `Flush::Finish` in `Writer::finish(&mut self)` to `D::Flush::finish()`.

[...]

PR #109, opened 2017-05-07
Implement distinct Flush types for Compress::compress vs Decompress::decompress

chrisvittal:
This implements distinct flush types (`FlushCompress` and `FlushDecompress`) for `Compress::compress` and `Decompress::decompress` as in #79. It also removes the old `Flush` type.

To keep the code generic in `read<R: BufRead, D: Ops>` and `Writer<W: Write, D: Ops>`, this adds an internal `Flush` trait for abstracting over the common values of the new flush types. This new `Flush` trait is used as a bound for a new associated type (also called `Flush`) for `Ops`.


src/mem.rs
```diff
@@ -114,7 +110,60 @@ pub enum Flush {
...
+    /// it should only be used when necessary. This will complete the current
+    /// deflate block and follow it with an empty stored block.
+    Sync = ffi::MZ_SYNC_FLUSH as isize,
+
+    /// Pending input is processed and pending output is flushed.
+    ///
+    /// The return value may indicate that the stream is not yet done and more
+    /// data has yet to be processed.
+    Finish = ffi::MZ_FINISH as isize
+}
```
alexcrichton: While we're at it, both of these flush enums are intended to be extensible in the sense that client code shouldn't match on them exhaustively. To accomplish this mind adding a `__Nonexhaustive` variant tagged with `#[doc(hidden)]` to both enums? That way it should hopefully send a signal of "we may extend this in the future".

alexcrichton:
👍 Thanks! I'll coordinate this with the 1.0 release so I won't merge this just yet, but rest assured I'll take it from here :)

src/mem.rs
```diff
@@ -91,18 +103,6 @@ pub enum Flush {
...
-    /// current block are held to be written as the next byte after the next
-    /// deflate block is completed.
-    ///
-    /// In this case the decompressor may not be provided enough bits at this
-    /// point in order to complete decompression of the data provided so far to
-    /// the compressor, it may need to wait for the next block to be emitted.
-    /// This is for advanced applications that need to control the emission of
-    /// deflate blocks.
-    Block = ffi::MZ_BLOCK as isize,
-
```
fstirlitz: Why was this removed? The commit message doesn't explain.

alexcrichton: @fstirlitz I believe it's because it's not actually used in the backend, but do you have a use case for it?

alexcrichton:
Indeed yeah thanks! I just need to allocate some time to go through the final steps of releasing 1.0 and then publish 1.0. I'll be sure to merge this before that though!

alexcrichton:
Alright I'm doing some prep work so I've merged, thanks so much again @chrisvittal for taking this on!

PR #120, opened 2017-08-28
Refactor the deflate, gz and zlib modules & rename types

sbstp:
I've re-factored the modules and types according to what was suggested in #75. Using `pub(crate)` in a few places, I was able to avoid having to create multiple layers of modules. Here's a summary of the changes:

* Split `deflate.rs`, `gz.rs` and `zlib.rs` into multiple files. Each module has a `mod.rs`, `bufread.rs`, `read.rs` and `write.rs`.
* The structs were renamed properly. `EncoderWriter` becomes `write::(Deflate|Gz|Zlib)Encoder`. `DecoderReader` becomes `read::(Deflate|Gz|Zlib)Decoder`. This was done for all the combinations.
* The `*Buf` structs are moved to their respective `bufread` submodule.
* The `*Reader` structs are moved to their respective `read` submodule.
* The `*Writer` structs are moved to their respective `write` submodule.
* Sprinkled a few `pub(crates)` here and there to keep the API unchanged.
* Grouped the struct declarations with their impl blocks in each file.
* The unit tests were moved to their respective `mod.rs`.


sbstp:
I can see in the travis log that I forgot to cleanup the imports when different features are used. On it.

alexcrichton:
Thanks so much for this! And wow that's quite the diff :)

I'll poke around with this and probably merge promptly after!

PR #130, opened 2017-10-23
Documentation - Mention rust backend & update miniz and flate info

oyvindln:
Note that there is now a rust back-end. (Granted a version with it hasn't been published yet so I don't know if this should be changed until that's ready.)

Point the miniz link to the new github page instead of the removed google code one.

Mention that the flate library is no longer part of the rust compiler, and fix the link so it doesn't point to the wrong library.
The libflate library on crates.io is a different library so I removed the lib prefix in the documentation here to avoid confusion.

Not sure if a link to flate is really needed but I did a simple replacement for now.

### DIFF
--- a/README.md
+++ b/README.md
@@ -6,8 +6,12 @@
 [![Documentation](https://docs.rs/flate2/badge.svg)](https://docs.rs/flate2)
 
 A streaming compression/decompression library for Rust. The underlying
-implementation by default uses [`miniz`](https://code.google.com/p/miniz/) but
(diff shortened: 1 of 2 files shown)


oyvindln:
Also, as miniz_oxide uses [`break` with value](https://github.com/rust-lang/rfcs/pull/1624) a fair bit, using the rust back-end requires at least rust 1.19. Should I put that in the readme as well?

alexcrichton:
Nah I think it's ok to not worry too much about the version requirements here, this otherwise looks great to me, thanks!

PR #135, opened 2017-11-28
Two little improvements

est31:
* Allow reading and setting the OS byte
* Derive Clone for GzHeader


est31:
I especially need the first commit, as it allows me to re-compress stuff that was compressed even with older versions of flate2-rs exactly the same way they were compressed.

About the second commit, it would allow me to make some of my code a bit nicer :).

alexcrichton:
Thanks!

PR #138, opened 2017-11-30
Emit a RFC 1952 compatible XFL flag again

### LINKED ISSUE #137: Breakage in XFL output
There is a regression in how commit 7e0390b51336b9737eefb10d1afc14d5e311fb9d has changed the XFL output. The diff:
```diff
-        header[8] = match lvl {
-            Compression::Best => 2,
-            Compression::Fast => 4,
-            _ => 0,
-        };
+        header[8] = lvl.0 as u8;
```

Previously, if you compressed stuff using `Compression::Best`, the XFL field would be set to 4. This is conformant with [RFC 1952](https://tools.ietf.org/html/rfc1952): 

```
         XFL (eXtra FLags)
            These flags are available for use by specific compression
            methods.  The "deflate" method (CM = 8) sets these flags as
            follows:

               XFL = 2 - compressor used maximum compression,
                         slowest algorithm
               XFL = 4 - compressor used fastest algorithm
```

It is also compatible with how unix tools treat that field.

With `flate2` 1.0, lvl.0 would be set to 9, so this behaviour was broken.

est31:
RFC 1952 specifies the following values
for the XFL flag:

* 2: maximum compression, slowest
* 4: fastest algorithm config

Commit 7e0390b51336b9737eefb10d1afc14d5e311fb9d has
changed our behaviour from conforming to that
to setting the XFL flag to 9 when compression
is best, and 0 when it is the fastest one.

We revert that change in order to have consistent
output with pre 1.0.0 versions of flate2 and also
to be consistent with RFC 1952 and the ecosystem
around it, e.g. unix tooling that assumes the
RFC specified layout for the XFL flag.

Fixes #137.

### DIFF
--- a/src/gz/mod.rs
+++ b/src/gz/mod.rs
@@ -236,7 +236,13 @@ impl GzBuilder {
         header[5] = (mtime >> 8) as u8;
         header[6] = (mtime >> 16) as u8;
         header[7] = (mtime >> 24) as u8;
-        header[8] = lvl.0 as u8;
+        header[8] = if lvl.0 >= Compression::best().0 {
+            2
+        } else if lvl.0 <= Compression::fast().0 {
+            4
+        } else {
+            0
+        };
 
         // Typically this byte indicates what OS the gz stream was created on,
         // but in an effort to have cross-platform reproducible streams just


est31:
I would love to have this added to a new 1.0.1 release. Especially, it would be really great to have this fix released before Cargo switches to flate2 1.0.

alexcrichton:
Thanks!

est31:
@alexcrichton thank you!

PR #141, opened 2017-12-22
Add gz::write::GzDecoder

vandenoever:
First stab at implementing gz::write::GzDecoder. It does not do crc check and the test hangs.

Should solve https://github.com/alexcrichton/flate2-rs/issues/113.

### DIFF
--- a/src/gz/bufread.rs
+++ b/src/gz/bufread.rs
@@ -34,7 +34,7 @@ fn read_le_u16<R: Read>(r: &mut R) -> io::Result<u16> {
     Ok((b[0] as u16) | ((b[1] as u16) << 8))
 }
 
-fn read_gz_header<R: Read>(r: &mut R) -> io::Result<GzHeader> {
+pub fn read_gz_header<R: Read>(r: &mut R) -> io::Result<GzHeader> {
     let mut crc_reader = CrcReader::new(r);
     let mut header = [0; 10];
     try!(crc_reader.read_exact(&mut header));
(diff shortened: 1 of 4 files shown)


src/gz/bufread.rs
```diff
@@ -34,7 +34,7 @@ fn read_le_u16<R: Read>(r: &mut R) -> io::Result<u16> {
     Ok((b[0] as u16) | ((b[1] as u16) << 8))
 }
 
-fn read_gz_header<R: Read>(r: &mut R) -> io::Result<GzHeader> {
+pub fn read_gz_header<R: Read>(r: &mut R) -> io::Result<GzHeader> {
```
KodrAus: We could make this `pub(super)` so it's only public in the `gz` module and not potentially available to other crates.

src/gz/write.rs
```diff
@@ -10,6 +10,8 @@ use super::GzBuilder;
 use {Compress, Compression};
 use crc::Crc;
 use zio;
+use super::super::deflate::write::DeflateDecoder;
```
KodrAus: Could this be `use deflate::write::DeflateDecoder`?

alexcrichton:
Thanks for this @vandenoever!

I think the main trickiness here is going to be receiving the header in bits and pieces instead of all at once. I don't think that we should return `Ok(0)` if the entire header wasn't written in one blob because sometimes the header could be a bit large. Instead we'll need to implement a state machine for decoding the header or otherwise buffering data until the entire header is decoded.

alexcrichton:
A bit late to the punch, but this was finished in https://github.com/alexcrichton/flate2-rs/pull/132

PR #145, opened 2018-01-11
Workaround to avoid having miniz_oxide_c_api export symbols that collide with miniz-sys

oyvindln:
For projects that have dependencies on both 0.2 and 1.0 versions of flate2 (particularly with zip-rs since it defaults to using the rust back-end). I didn't manage to trigger this myself, but someone else did: https://gitter.im/rust-lang/rust?at=5a57cc7183152df26d5d5cfc

This should hopefully avoid this for now. The c api wrapping was mainly used for ease of integration, ideally we would avoid it entirely for the rust backend, though that would be a much bigger change.


alexcrichton:
Thanks!

PR #174, opened 2018-11-19
Add license symlinks to flate2-crc

erickt:
Presumably these use the same text as the parent flate2 project. This
patch makes sure the crates.io artifact includes the license files.


erickt:
Thanks for the quick merge! would it be possible to get a release cut with this so the crates.io artifact includes the license?

alexcrichton:
Sure thing, I've now published this

erickt:
Thanks Alex!

PR #180, opened 2018-12-07
asyncread + asyncwrite trait impl on GzEncoder and GzDecoder

yanilov:
WIP - need to add tests. help appreciated.

### DIFF
--- a/src/gz/bufread.rs
+++ b/src/gz/bufread.rs
@@ -3,6 +3,11 @@ use std::io;
 use std::io::prelude::*;
 use std::mem;
 
+#[cfg(feature = "tokio")]
+use futures::Poll;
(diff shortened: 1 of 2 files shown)


alexcrichton:
Oops sorry for the delay here! IIRC these weren't added historically because the actual implementations weren't ready for async usage, something about them required something to be synchronous. I sort of forget though which aspect that was, and it may not be true any more...

In any case some tests would be good to add (especially some that limited reads to succeed with maximally one byte at a time), and then this seems fine to add!

yanilov:
You are correct! I experimented with it and I'm getting different results in an asynchronous flow.
I'm closing this PR but keeping the issue open.

PR #183, opened 2019-01-18
Updated to new tokio API

dodomorandi:
This should be helpful to introduce new async features, and maybe it
could ease the migration to the currently unstable futures.

I had to change the name of the feature from 'tokio' to 'async', because we cannot have crates and features with the same name.

Any feedback is appreciated!

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -2,7 +2,7 @@
 
 name = "flate2"
 authors = ["Alex Crichton <alex@alexcrichton.com>"]
-version = "1.1.0"
+version = "1.0.6"
 license = "MIT/Apache-2.0"
 readme = "README.md"
 keywords = ["gzip", "flate", "zlib", "encoding"]
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -2,7 +2,7 @@
 
 name = "flate2"
 authors = ["Alex Crichton <alex@alexcrichton.com>"]
-version = "1.0.6"
+version = "1.1.0"
 license = "MIT/Apache-2.0"
 readme = "README.md"
 keywords = ["gzip", "flate", "zlib", "encoding"]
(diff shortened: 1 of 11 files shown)


alexcrichton:
Thanks for this! I'd be somewhat wary though about changing names of the feautres, but `tokio` is a facade I think so perhaps smaller components could be used?

dodomorandi:
Ups, you are totally right! :sweat_smile: I will revert back the feature name and I will use the sub-crates of `tokio` instead of the whole facade.

alexcrichton:
Thanks! I think now this PR is largely just updating some idioms, having stricter requirements on the tokio-io and futures crates, and updating tests, right? If so I think it's safe to skip the version bump here as I don't think it'll be critical to get out to crates.io

dodomorandi:
I was not sure about the version bump, because of the old `tokio-core' and a few differences with the current environment. At the end I bumped just because if we forgot about it and merged, if could be a mess.

Surely the `TcpStream` API changed and both `tokio::core` and the `try_nb` macro are not available in newer versions. What do you recommend to evaluate if those are breaking changes for existent projects?

alexcrichton:
I think that tokio-core is only used during this crate's own tests and not used externally though, so the changes for users on crates.io are just internal rewrites. Otherwise the crate both before and after should compile with the same set of compilers and crates.

dodomorandi:
Perfect! If it is ok, I can change the version bump to 1.0.7, or I can revert it back to 1.0.6 if you would rather to include some other changes before the new version.

alexcrichton:
Since this is now largely a cleanup PR I'd prefer to leave it at 1.0.6 and we can let these changes ride to crates.io on the next bump

alexcrichton:
Thanks!

PR #185, opened 2019-02-14
Async GzDecoder

### LINKED ISSUE #179: missing AsyncRead + AsyncWrite for gzip implementations
GzEncoder and GzDecoder (+ bufread variants) are missing tokio_io's AsyncRead and AsyncWrite  trait implementations

quininer:
This PR implements async gzip header parsing. It will fixed #179 .

* ~~It changed some behavior of `GzDecoder` and was a breaking change.~~

### DIFF
--- a/src/gz/bufread.rs
+++ b/src/gz/bufread.rs
@@ -329,14 +329,9 @@ impl<'a, T: Read> Read for Buffer<'a, T> {
     fn read(&mut self, buf: &mut [u8]) -> io::Result<usize> {
         let mut len = self.buf.read(buf)?;
         if buf.len() > len {
-            match self.reader.read(&mut buf[len..])? {
-                // eof
-                0 => return Err(bad_header()),
-                len2 => {
-                    self.buf.get_mut().get_mut().extend_from_slice(&buf[len..][..len2]);
-                    len += len2;
-                }
-            }
+            let len2 = self.reader.read(&mut buf[len..])?;
+            self.buf.get_mut().get_mut().extend_from_slice(&buf[len..][..len2]);
+            len += len2;
         }
         Ok(len)
     }
@@ -360,7 +355,6 @@ impl<R: BufRead> GzDecoder<R> {
                 GzState::Body
             },
             Err(ref err) if io::ErrorKind::WouldBlock == err.kind()
-                || io::ErrorKind::UnexpectedEof == err.kind()
                 => GzState::Header(buf),
             Err(err) => GzState::Err(err)
         };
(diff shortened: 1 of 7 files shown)


src/gz/mod.rs
```diff
@@ -316,11 +316,11 @@ mod tests {
             .extra(vec![0, 1, 2, 3])
             .read(&r[..], Compression::default());
         let mut d = read::GzDecoder::new(e);
-        assert_eq!(d.header().unwrap().filename(), Some(&b"foo.rs"[..]));
-        assert_eq!(d.header().unwrap().comment(), Some(&b"bar"[..]));
-        assert_eq!(d.header().unwrap().extra(), Some(&b"\x00\x01\x02\x03"[..]));
         let mut res = Vec::new();
         d.read_to_end(&mut res).unwrap();
+        assert_eq!(d.header().unwrap().filename(), Some(&b"foo.rs"[..])); // must read first
```
quininer: Usually in async code, we don't want to read IO before poll. The new behavior does not allow us to get gzip header before read.

I don't know if it's worth release 2.0 for this.

alexcrichton:
Thanks for the PR! I've done a quick skim over this and was wondering if a few high-level points could be knocked out before doing a deeper review:

* Instead of having an enum of states we cycle between, could we just buffer data internally until the whole header has been read? In theory the header is quite small and won't take long, even over async connections, to wait for it to all arrive.
* Could the `new2` functions be avoided? Ideally this doesn't affect the public API, only adding a few trait impls as necessary to publicly exposed types.

quininer:
* I have tried to use `BufRead` instead of a state machine for parsing, but `GzWriter` doesn't provide `BufRead` for this, so I switched to a state machine instead. If you prefer `BufRead`, I am willing to make changes.
* In an async context, we usually don't want to consume read before polled. This caused us to not get the header immediately, which is a breaking change.

alexcrichton:
Right yeah I don't think we need to deal with `BufRead`, just read data into a local buffer until the full header is available and then we parse it all in one go.

I'm not sure I understand your comments about `new2` though, can you clarify?

quininer:
I may have thought wrong before. :(

quininer:
ping? @alexcrichton

src/gz/bufread.rs
```diff
@@ -280,108 +298,213 @@ impl<R: BufRead + Write> Write for GzEncoder<R> {
...
+    Body,
+    Finished(usize, [u8; 8]),
+    Err(io::Error),
+    End
+}
+
+struct Buffer<'a, T> {
+    buf: io::Take<io::Cursor<&'a mut Vec<u8>>>,
+    reader: &'a mut T
+}
```
alexcrichton: I think this can probably be replaced with `Chain<&[u8], T>` perhaps? Or rather if you use `io::Read::chain` where the first part is the `&[u8]` slice and the second is `&mut T`, I think this adapter may not be needed?

src/gz/bufread.rs
```diff
@@ -280,108 +298,213 @@ impl<R: BufRead + Write> Write for GzEncoder<R> {
...
+            match inner {
+                GzState::Header(buf) => {
+                    let mut reader = Buffer::new(buf, reader.get_mut().get_mut());
+                    match read_gz_header(&mut reader) {
+                        Ok(hdr) => {
+                            *header = Some(hdr);
+                            next = Next::Body;
+                        },
+                        Err(ref err) if io::ErrorKind::WouldBlock == err.kind()
+                            || io::ErrorKind::UnexpectedEof == err.kind()
```
alexcrichton: Out of curiosity, how come `UnexpectedEof` is here in addition to `WouldBlock`? I'd expect that `UnexpectedEof` would be a "completely fatal" error we'd want to return rather than handle

src/gz/bufread.rs
```diff
@@ -280,108 +298,213 @@ impl<R: BufRead + Write> Write for GzEncoder<R> {
...
-        match self.header {
-            None => return Ok(0), // error already returned,
-            Some(Ok(_)) => {}
-            Some(Err(_)) => match self.header.take().unwrap() {
-                Ok(_) => panic!(),
-                Err(e) => return Err(e),
-            },
+        let GzDecoder { inner, header, reader, multi } = self;
+
+        enum Next {
```
alexcrichton: Having `enum Next` here makes it a bit hard to follow the control flow, especially `return Err` for wouldblock errors and a fall-through for other errors. Would it be possible to replace the current state with `None` on each iteration? That'd ideally remove the need for this `Next` and would also enable `?` to be used more uniformly in a few places.

src/gz/bufread.rs
quininer: This adapter is different from `Chain`, which copies data into `buf` at read time so that it can be reused when an error occurs.

I think that a simple `Chain<&[u8], T>` can't do this, we still need a `Buffer<Chain<&[u8], T>>`.

src/gz/bufread.rs
quininer: You are right... `UnexpectedEof` should not appear.

src/gz/bufread.rs
quininer: This is not as good as it is supposed to be, because we still need to set inner to correct state before throwing `WouldBlock`.

src/gz/bufread.rs
alexcrichton: Ah ok, makes sense!

alexcrichton:
Ok looks good to me, thanks!

quininer:
Can you release a new version for it?

alexcrichton:
Published!

PR #196, opened 2019-06-07
Add `Builder` to configure `Compress`/`Decompress`.

twittner:
To allow more control over certain parameters (e.g. the window size) and to not break the existing API, a configuration builder is added in this PR. The control is needed for example to properly implement the permessage-deflate extension of [RFC 7692](https://tools.ietf.org/html/rfc7692#section-7).

I chose to restrict the window bits to the range of 9 ..= 15 (instead of 8 ..= 15), because of zlib's recommendation in [`deflateInit2`](https://www.zlib.net/manual.html#Advanced):

> For the current implementation of `deflate()`, a `windowBits` value of 8 (a window size of 256 bytes) is not supported. As a result, a request for 8 will result in 9 (a 512-byte window). In that case, providing 8 to `inflateInit2()` will result in an error when the *zlib* header with 9 is checked against the initialization of `inflate()`. The remedy is to not use 8 with `deflateInit2()` with this initialization, or at least in that case use 9 with `inflateInit2()`.

### DIFF
--- a/src/mem.rs
+++ b/src/mem.rs
@@ -191,63 +191,45 @@ pub enum Status {
     StreamEnd,
 }
 
-/// Configuration builder for [`Compress`] and [`Decompress`].
-#[derive(Debug)]
-pub struct Builder<T> {
-    window_bits: c_int,
-    zlib_header: bool,
-    level: Compression,
-    _marker: std::marker::PhantomData<T>
-}
-
-impl<T> Builder<T> {
-    fn new() -> Self {
-        Builder {
-            window_bits: ffi::MZ_DEFAULT_WINDOW_BITS,
-            zlib_header: false,
-            level: Compression::fast(),
-            _marker: std::marker::PhantomData
-        }
+impl Compress {
+    /// Creates a new object ready for compressing data that it's given.
+    ///
+    /// The `level` argument here indicates what level of compression is going
+    /// to be performed, and the `zlib_header` argument indicates whether the
+    /// output data should have a zlib header or not.
+    pub fn new(level: Compression, zlib_header: bool) -> Compress {
+        Compress::make(level, zlib_header, ffi::MZ_DEFAULT_WINDOW_BITS as u8)
     }
 
-    /// Set the maximum window bits to use.
+    /// Creates a new object ready for compressing data that it's given.
+    ///
+    /// The `level` argument here indicates what level of compression is going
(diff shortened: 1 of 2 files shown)


src/mem.rs
```diff
@@ -223,6 +262,49 @@ impl Compress {
...
+            let mut state = ffi::StreamWrapper::default();
+            let ret = ffi::mz_inflateInit2(
+                &mut *state,
+                if self.zlib_header {
+                    self.window_bits
+                } else {
+                    -self.window_bits
+                },
+            );
+            debug_assert_eq!(ret, 0);
```
oyvindln: Woudln't it be best to have the assert in release too? Otherwise there will be issues if inflateInit fails.

oyvindln: Which it might since [miniz](https://github.com/alexcrichton/flate2-rs/blob/master/miniz-sys/miniz.c#L195) and miniz_oxide only support one window size

twittner: Thank you! I applied your suggestion and also changed `Builder::window_bits` to only be available for the `zlib` feature.

alexcrichton:
Thanks for the PR! I think though in this case having a builder may be a bit overkill, could a new constructor taking the window bits parameter be added?

twittner:
I am fine with using an additional constructor instead of a builder. However I am not sure how to name it. What would be your preference?

alexcrichton:
I'd be fine with something like `new_with_window_bits` or something like that?

alexcrichton:
Thanks! One thing I may have missed though, how come these are only available on zlib? 

(could the documentation be updated to reflect that as well?)

twittner:
> [..] how come these are only available on zlib?

@oyvindln pointed it out [above](https://github.com/alexcrichton/flate2-rs/pull/196#discussion_r291641215), linking to https://github.com/alexcrichton/flate2-rs/blob/master/miniz-sys/miniz.c#L195

> could the documentation be updated to reflect that as well?

Sure, I'll add a note.

alexcrichton:
Aha that must be why I didn't expose this originally as well!

Thanks!

twittner:
@alexcrichton: Thanks for merging this! Do you plan to publish a new version to crates.io anytime soon?

alexcrichton:
Sure thing, published now!

twittner:
Thanks a lot!

PR #198, opened 2019-06-14
Expose a few more functions

Keats:
I'm building a streaming compression/decompression system with flate2-rs (eg from a multipart upload where you get some chunks once in a while and do not want to keep them in memory once compressed/decompressed) and having those functions public would avoid copying quite a bit of code.
I can't use a fork because of some miniz-sys conflict with some sub-dependencies of actix-web apparently?

### DIFF
--- a/src/crc.rs
+++ b/src/crc.rs
@@ -60,6 +60,21 @@ impl Crc {
         self.amt += additional_crc.amt;
         self.hasher.combine(&additional_crc.hasher);
     }
+
+    /// Computes the 8 bytes CRC
+    pub fn to_bytes(&self) -> [u8; 8] {
+        let (sum, amt) = (self.sum() as u32, self.amount());
+        [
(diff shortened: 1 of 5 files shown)


alexcrichton:
Thanks for the PR! I'm somewhat uncomfortable just making internal details public though. Most of these don't have great interfaces in that they're not super ergonomic or make too many assumptions about return arguments and such. For example they might allocate too much or not be flexible enough on inputs.

I would personally prefer to put more thought into the design of these APIs instead of just exporting what happens to already be there.

Keats:
No worries, I figured it might be the case. I'll just copy/paste what I need then. I do think the API are pretty good tbh

alexcrichton:
Ok in that case I'm gonna go ahead and close this. I think it's fine to stabilize more functionality but I'd prefer that it would be done deliberately rather than simply exposing what's already there.

PR #202, opened 2019-08-07
[WIP] Use miniz_oxide directly in rust backend

### LINKED ISSUE #181: Directly use miniz_oxide
In Debian we're patching away usage of miniz and miniz_oxide because it's simpler to just use the system zlib. The "rust_backend" feature actually just points to zlib.

If flate2/rust_backend uses miniz_oxide directly (rather than through miniz_oxide_c_api) then it would be easier to have a "real" rust_backend actually in Debian.

oyvindln:
This avoids the C API and it's unsafeness entirely.

Can probably further simplify a bit by altering the Stream struct and direction trait StreamWrapper doesn't have to be an enum, neither are public.

Needs some thorough testing.

fixes #181

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -20,7 +20,7 @@ streams.
 members = ['systest']
 
 [dependencies]
-libc = "0.2"
+libc = { version = "0.2", optional = true }
 miniz-sys = { path = "miniz-sys", version = "0.1.11", optional = true }
 libz-sys = { version = "1.0", optional = true }
 tokio-io = { version = "0.1.11", optional = true }
@@ -37,8 +37,8 @@ tokio-threadpool = "0.1.10"
 futures = "0.1"
 
 [features]
-default = ["miniz-sys"]
-zlib = ["libz-sys"]
+default = ["miniz-sys", "libc"]
+zlib = ["libz-sys", "libc"]
 rust_backend = ["miniz_oxide"]
 tokio = ["tokio-io", "futures"]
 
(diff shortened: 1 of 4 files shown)


alexcrichton:
Thanks for this! One thing I'd like to do here is to reduce the number of `#[cfg]` necessary. Would it be possible to push more of the `#[cfg]` into the `src/ffi.rs` module itself? That way the backend-specific implementations could all be located there and ideally `src/mem.rs` is free from `#[cfg]` (largely at least)

oyvindln:
I'll look into it.

oyvindln:
Right, pushed all of the back-end specific stuff into ``ffi.rs``. Only ones that remain are the zlib-specific public functions.

Is there some specific reason the zlib ```mz_stream``` is boxed, while the miniz one is not? Otherwise they could share that code.

alexcrichton:
That's actually probably a bug, want to go ahead and box both of them up? I think there's also an error on nightly to handle?

oyvindln:
I'll box both then.

The nightly error may actually be something that's in the original code, the lint was added recently, and the original code uses ```mem::zeroed()``` here too. I suspect it's the ```extern fn``` fields that the compiler does not want to assume zero is a valid value for.

alexcrichton:
Hm yeah so I think I messed that up when I wrote libz-sys.

If you wouldn't mind adding `#[allow(invalid_value)]` for now I can try to clean this up after this lands.

alexcrichton:
👍

PR #216, opened 2019-10-01
Switch to the Rust backend by default

### LINKED ISSUE #215: Use miniz_oxide by default
It seems miniz_oxide backend has matured enough to be the default:

1. Now that `miniz_oxide` backend contains no unsafe code and calls to it no longer go through a C API, it is strictly safer than the default miniz backend.
1. Performance is about the same (oxide decompression is faster, compression is slower).
1. miniz_oxide has received plenty of testing since the popular `reqwest` library uses it exclusively.

Is there anything else needed before flate2 can switch to miniz_oxide by default?

alexcrichton:
cc @oyvindln and @Frommi, I'd want to make sure y'all are ok with this and if you're up for it. This is likely to give the crate quite a bit of exposure, and I'm also curious if y'all think it's ready for this? Things are looking good to me, and I think it should be fine to do this in a point release!

Closes #215


oyvindln:
Sounds good to me.

PR #221, opened 2019-11-01
cargo: update minimum versions

mathstuf:
crc32fast has dependency specifications which are compatible and cfg-if
0.1.6 is the first to include APIs used in the ffi module.


mathstuf:
Would it be possible to get a release with this included made?

alexcrichton:
Sure thing, done!

PR #226, opened 2019-12-15
Remove deprecated Error::description

kornelski:


kornelski:
Rustfmt failed the build. I don't like rustfmt, so I'm not fixing that.

PR #228, opened 2020-02-05
Support cloudfire optimized version of zlib as a backend

arthurprs:
I had a use case that needed a faster gzip encoder. Is this something interesting to push upstream? 

Some brief testing showed that compression at level 6 is twice as fast as the default backend.

### DIFF
--- a/.github/workflows/main.yml
+++ b/.github/workflows/main.yml
@@ -39,6 +39,7 @@ jobs:
     - run: cargo test --features zlib
     - run: cargo test --features miniz-sys
     - run: cargo test --features zlib --no-default-features
+    - run: cargo test --features cloudflare_zlib --no-default-features
     - run: cargo test --features miniz-sys --no-default-features
     - run: cargo test --features tokio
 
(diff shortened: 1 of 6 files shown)


alexcrichton:
Nice! 

Mind adding this to CI as well to get tested?

arthurprs:
I'm unsure why but CI says it doesn't compile under mingw.

alexcrichton:
It looks like inline assembly is used and that probably isn't compatible with the MinGW toolchain. Perhaps document that it doesn't work on MinGW and we can configure in the code to ignore the feature on MinGW?

arthurprs:
I modified the build to skip this feature on mingw and added a small note to the README.

alexcrichton:
Thanks!

RReverser:
@arthurprs It looks like there is a bunch of `feature = "zlib"` conditions in https://github.com/alexcrichton/flate2-rs/blob/master/src/mem.rs that weren't extended to include `cloudflare_zlib`.

Is that intentional or an accidental omission?

arthurprs:
@RReverser good eye. I overlooked those.

RReverser:
Just made a PR above.

PR #240, opened 2020-05-20
feat: allow full range of window bits for zlib

ryanvade:
According to https://github.com/madler/zlib/blob/cacf7f1d4e3d44d871b605da3b647f07d718623f/zlib.h#L839-L870 the Window Bits parameter for `inflateInit2` should support values in the range of -15 to 47. This PR allows for those ranges in the C implementation.

### DIFF
--- a/src/ffi/c.rs
+++ b/src/ffi/c.rs
@@ -163,10 +163,10 @@ pub struct Inflate {
 }
 
 impl InflateBackend for Inflate {
-    fn make(zlib_header: bool, window_bits: u8) -> Self {
+    fn make(zlib_header: bool, window_bits: i32) -> Self {
         assert!(
-            window_bits > 8 && window_bits < 16,
-            "window_bits must be within 9 ..= 15"
+            window_bits > -16 && window_bits < 48,
+            "window_bits must be within -15 ..= 47"
         );
         unsafe {
             let mut state = StreamWrapper::default();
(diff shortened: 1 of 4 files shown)


oyvindln:
May cause issues since negative values are intended to be used to indicate the zlib header (which there is another parameter for already.)

ryanvade:
> May cause issues since negative values are intended to be used to indicate the zlib header and >15 means gzip header. There are already other structures for that purpose in flate2.

True, but right now flipping the zlib_header flag only changes the range from 8..15 to -15..-8 while zlib supports values > 15. Perhaps removing the zlib_header flag altogether would be a better solution? Right now I'm trying to solve https://github.com/alexcrichton/flate2-rs/issues/239

alexcrichton:
Could you be sure to add a test for this as well?

joshtriplett:
Reading through this carefully, I don't think it makes sense to allow passing other values for window bits to the existing objects, because that would mean it'd be possible to use (for instance) `GzDecoder` but pass a value for window bits that implies a zlib stream or a raw DEFLATE stream.

It *does* make sense to support autodetection of zlib vs gzip, but for that, I think flate2 should have a separate `AutoDecoder` or similar.

coolreader18:
I'm looking into this to easily support Python's window bits options ("+25 to +31 = 16 + (9 to 15): Uses the low 4 bits of the value as the window size logarithm, while including a basic gzip header and trailing checksum in the output.") in RustPython; do you think it would make sense to add a `cfg(feature = "any_zlib")` method like `Decompress::new_gzip(window_bits: u8)` that would pass `window_bits + 16` to `inflateInit`?

coolreader18:
Or I suppose it would have to be `cfg(feature = "zlib")`, miniz doesn't support gzip headers. Alternatively, maybe there could be some sort of `from_raw` function to create a `Compress`/`Decompress` from a C `z_stream`?

alexcrichton:
Yes I think this will at least need to be a zlib-specific feature, but otherwise seems fine by me to expose.

PR #242, opened 2020-06-22
Reset mem_level to DEF_MEM_LEVEL. Close #241

### LINKED ISSUE #241: ZlibEncoder uses different compress option from zlib when using zlib feature
When compressing data, the result flate2 produced is different from zlib(the C lib). The reason is flate2 uses a different argument to call `deflateInit2_`.

https://github.com/alexcrichton/flate2-rs/blob/962930cb3a1b1c50b710c9f4749f1686365eb70a/src/ffi/c.rs#L267-L277
The `mem_level` is setted to 9 instead 8 (`DEF_MEM_LEVEL` zlib used).
https://github.com/madler/zlib/blob/cacf7f1d4e3d44d871b605da3b647f07d718623f/zutil.h#L66
https://github.com/madler/zlib/blob/cacf7f1d4e3d44d871b605da3b647f07d718623f/deflate.c#L234-L235

I think a reason is that miniz use 9 as it's default mem_level, but that makes no sense.
https://github.com/Frommi/miniz_oxide/blob/e91d8fca64617b09f236e4b939f4971a891abe27/miniz/miniz.c#L185

Reset that argument to `DEF_MEM_LEVEL` is useful, because when I using flate2's zlib feature I want to get the same result as using zlib in C.

constfold:
Fixes #241

### DIFF
--- a/src/ffi/c.rs
+++ b/src/ffi/c.rs
@@ -273,7 +273,7 @@ impl DeflateBackend for Deflate {
                 } else {
                     -(window_bits as c_int)
                 },
-                9,
+                8,
                 MZ_DEFAULT_STRATEGY,
             );
             assert_eq!(ret, 0);


alexcrichton:
Sounds good to me, thanks!

PR #247, opened 2020-08-14
Cargo.toml: rewrite description, no longer "bindings to miniz.c"

joshtriplett:
flate2 does not just provide "bindings to miniz.c", and in fact doesn't
use miniz.c by default. Rewrite the description for clarity.


alexcrichton:
👍

PR #248, opened 2020-08-18
Update to libz-sys 1.1.0, and support use with zlib-ng

joshtriplett:
libz-sys 1.1.0 introduces optional support for zlib-ng, if the default
`stock-zlib` feature is disabled and the `zlib-ng` feature is enabled.
Modify the dependency on libz-sys to disable the `stock-zlib` feature,
which allows crates using flate2 to opt into zlib-ng if they wish. Add a
`zlib-ng-compat` feature to flate2, to opt into zlib-ng explicitly.

Update the README to document this, and to describe some of the
tradeoffs between zlib implementations.

### DIFF
--- a/.github/workflows/main.yml
+++ b/.github/workflows/main.yml
@@ -39,6 +39,7 @@ jobs:
     - run: cargo test --features zlib
     - run: cargo test --features miniz-sys
     - run: cargo test --features zlib --no-default-features
+    - run: cargo test --features zlib-ng-compat --no-default-features
     - run: cargo test --features cloudflare_zlib --no-default-features
       if: matrix.build != 'mingw'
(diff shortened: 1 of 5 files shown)


joshtriplett:
(Updated: ran rustfmt.)

joshtriplett:
I used the feature name `zlib-ng-compat`, to reserve the feature name `zlib-ng` for a potential future change that enables use of the standalone zlib-ng library (built without zlib-compat mode).

alexcrichton:
Nice! Could this also add testing to CI which builds this on all the platforms and tests it?

joshtriplett:
@alexcrichton Done, and it passes.

PR #252, opened 2020-10-01
Make `Compression::{new, none, fast, best}` const fn

notviri:
Hey, I'm not too familiar with the internals of this compression algorithm but this seems to just be a newtype over an integer? It's the only type I can't construct const-ly in a project, so maybe it could be changed, I hope!

(I don't exactly see a MSRV but I'm pretty sure `const fn` with an integer like this is long supported)


alexcrichton:
Thanks!

PR #270, opened 2021-04-21
Use the error message from z_stream.msg for [De]CompressError

coolreader18:
Not sure if I should put a `#[cfg]` on the `msg` field of the errors for only when zlib is used, since the other backends don't produce errors, or if that would be needlessly complicated.

### DIFF
--- a/src/ffi/c.rs
+++ b/src/ffi/c.rs
@@ -2,6 +2,7 @@
 use std::alloc::{self, Layout};
 use std::cmp;
 use std::convert::TryFrom;
+use std::ffi;
 use std::fmt;
 use std::marker;
 use std::ops::{Deref, DerefMut};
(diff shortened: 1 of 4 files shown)


src/ffi/c.rs
```diff
@@ -138,6 +139,18 @@ pub struct Stream<D: Direction> {
     pub _marker: marker::PhantomData<D>,
 }
 
+impl<D: Direction> Stream<D> {
+    pub fn msg(&self) -> Option<&'static str> {
```
alexcrichton, line 143: To confirm, are you sure that all C backends use this as `&'static str` in that it always has a static lifetime?

If not I think the error will need to start containing a `String`?

alexcrichton (commented):
Looks reasonable to me, thanks! I didn't actually realize this error information was present...


Could you add a test that the message is present when we expect it to be present?

src/ffi/c.rs
coolreader18: Well zlib does (and miniz just never changes `msg` from NULL), and I figure that if we eventually need to store more dynamic error messages we can change it to Cow. `De/CompressError.message` only returns a `&str`, not a `&'static str`, so I figure this should be fine.

coolreader18:
I did end up changing the error types so that there's no overhead to this when using miniz/miniz_oxide, I don't think it's too bad. Also, I noticed that some of the cfgs from a previous contribution of mine (`[De]Compress::new_gzip`) as well as misc stuff in `ffi::c` were inaccurate/verbose, so I tidied that up as well. Let me know if you'd rather I split that out to a separate PR.

alexcrichton:
Nah this looks good, thanks!

PR #277, opened 2021-07-11
Renamed corrupt test gzip file

billallen256:
The corrupt gzip file used in the tests prevents some systems from mirroring this crate because the system's antivirus software rejects corrupt files that it cannot scan.


alexcrichton:
Thanks!

PR #278, opened 2021-07-19
Avoid quadratic complexity in GzDecoder

catenacyber:
Quadratic complexity happens when the underlying reader
supplies bytes one at a time and would block.
And if header flag FNAME is set and we never have a null
byte to end the filename, read_gz_header keeps reading
all of the partial filename and computing the CRC...

The solution is to keep a more complex structure than a
simple buffer for what has already been parsed in the
header (as is done in C zlib)

@alexcrichton I will put some comments in the code for your review

### DIFF
--- a/src/gz/bufread.rs
+++ b/src/gz/bufread.rs
@@ -10,7 +10,7 @@ use tokio_io::{AsyncRead, AsyncWrite};
 
 use super::{GzBuilder, GzHeader};
 use super::{FCOMMENT, FEXTRA, FHCRC, FNAME};
-use crate::crc::CrcReader;
+use crate::crc::{Crc, CrcReader};
 use crate::deflate;
 use crate::Compression;
 
@@ -38,6 +38,90 @@ fn read_le_u16<R: Read>(r: &mut R) -> io::Result<u16> {
     let mut b = [0; 2];
     r.read_exact(&mut b)?;
     Ok((b[0] as u16) | ((b[1] as u16) << 8))
+}
+
+fn read_le_u16_once<R: Read>(r: &mut Buffer<R>) -> io::Result<u16> {
+    let mut b = [0; 2];
+    r.read_once(&mut b)?;
+    Ok((b[0] as u16) | ((b[1] as u16) << 8))
+}
+
+fn read_gz_header_part<R: Read>(r: &mut Buffer<R>) -> io::Result<()> {
+    loop {
+        match r.part.state {
+            GzHeaderParsingState::GzHeaderParsingStart => {
+                let mut header = [0; 10];
+                r.read_once(&mut header)?;
+                r.part.buf.truncate(0);
+
+                if header[0] != 0x1f || header[1] != 0x8b {
+                    return Err(bad_header());
+                }
+                if header[2] != 8 {
+                    return Err(bad_header());
+                }
+
+                r.part.flg = header[3];
+                r.part.mtime = ((header[4] as u32) << 0)
+                    | ((header[5] as u32) << 8)
+                    | ((header[6] as u32) << 16)
+                    | ((header[7] as u32) << 24);
+                let _xfl = header[8];
+                r.part.os = header[9];
+                r.part.state = GzHeaderParsingState::GzHeaderParsingXlen;
+            }
+            GzHeaderParsingState::GzHeaderParsingXlen => {
+                if r.part.flg & FEXTRA != 0 {
+                    r.part.xlen = read_le_u16_once(r)?;
+                }
+                r.part.state = GzHeaderParsingState::GzHeaderParsingExtra;
+            }
+            GzHeaderParsingState::GzHeaderParsingExtra => {
+                if r.part.flg & FEXTRA != 0 {
+                    let mut extra = vec![0; r.part.xlen as usize];
+                    r.read_once(&mut extra)?;
+                    r.part.extra = Some(extra);
+                }
+                r.part.state = GzHeaderParsingState::GzHeaderParsingFilename;
+            }
+            GzHeaderParsingState::GzHeaderParsingFilename => {
+                if r.part.flg & FNAME != 0 {
+                    for byte in r.reader.bytes() {
+                        let byte = byte?;
+                        r.crc.update(&[byte]);
+                        if byte == 0 {
+                            break;
+                        }
+                        r.part.filename.push(byte);
+                    }
+                }
+                r.part.state = GzHeaderParsingState::GzHeaderParsingComment;
+            }
+            GzHeaderParsingState::GzHeaderParsingComment => {
+                if r.part.flg & FCOMMENT != 0 {
+                    for byte in r.reader.bytes() {
+                        let byte = byte?;
+                        r.crc.update(&[byte]);
+                        if byte == 0 {
+                            break;
+                        }
+                        r.part.comment.push(byte);
+                    }
+                }
+                r.part.state = GzHeaderParsingState::GzHeaderParsingCrc;
+            }
+            GzHeaderParsingState::GzHeaderParsingCrc => {
+                if r.part.flg & FHCRC != 0 {
+                    let stored_crc = read_le_u16_once(r)?;
+                    let calced_crc = r.crc.sum() as u16;
+                    if stored_crc != calced_crc {
+                        return Err(corrupt());
+                    }
+                }
+                return Ok(());
+            }
+        }
+    }
 }
 
 pub(crate) fn read_gz_header<R: Read>(r: &mut R) -> io::Result<GzHeader> {
@@ -305,8 +389,60 @@ pub struct GzDecoder<R> {
 }
 
 #[derive(Debug)]
+pub enum GzHeaderParsingState {
+    GzHeaderParsingStart = 0,
+    GzHeaderParsingXlen = 1,
+    GzHeaderParsingExtra = 2,
+    GzHeaderParsingFilename = 3,
+    GzHeaderParsingComment = 4,
+    GzHeaderParsingCrc = 5,
+}
+
+#[derive(Debug)]
+pub struct GzHeaderPartial {
+    buf: Vec<u8>,
+
+    state: GzHeaderParsingState,
+    flg: u8,
+    os: u8,
+    xlen: u16,
+    mtime: u32,
+    extra: Option<Vec<u8>>,
+    filename: Vec<u8>,
+    comment: Vec<u8>,
+}
+
+impl GzHeaderPartial {
+    fn new() -> GzHeaderPartial {
+        GzHeaderPartial {
+            buf : Vec::with_capacity(10), // minimum header length
+            state : GzHeaderParsingState::GzHeaderParsingStart,
+            flg : 0,
+            os : 0,
+            xlen : 0,
+            mtime: 0,
+            extra: None,
+            filename: Vec::new(),
+            comment: Vec::new(),
+        }
+    }
+
+    pub fn take_header(self) -> GzHeader {
+        let filename = if self.flg & FNAME != 0 {Some(self.filename)}  else {None};
+        let comment = if self.flg & FCOMMENT != 0 {Some(self.comment)}  else {None};
+        return GzHeader {
+            extra: self.extra,
+            filename: filename,
+            comment: comment,
+            operating_system: self.os,
+            mtime: self.mtime,
+        };
+    }
+}
+
+#[derive(Debug)]
 enum GzState {
-    Header(Vec<u8>),
+    Header(GzHeaderPartial),
     Body,
     Finished(usize, [u8; 8]),
     Err(io::Error),
@@ -317,19 +453,21 @@ enum GzState {
 /// further data from `reader`. This will also buffer all data read from
 /// `reader` into `buf` for reuse on a further call.
 struct Buffer<'a, T: 'a> {
-    buf: &'a mut Vec<u8>,
+    part: &'a mut GzHeaderPartial,
     buf_cur: usize,
     buf_max: usize,
+    crc: Crc,
     reader: &'a mut T,
 }
 
 impl<'a, T> Buffer<'a, T> {
-    fn new(buf: &'a mut Vec<u8>, reader: &'a mut T) -> Buffer<'a, T> {
+    fn new(part: &'a mut GzHeaderPartial, reader: &'a mut T) -> Buffer<'a, T> {
         Buffer {
             reader,
             buf_cur: 0,
-            buf_max: buf.len(),
-            buf,
+            buf_max: part.buf.len(),
+            crc: Crc::new(),
+            part,
         }
     }
 }
@@ -338,13 +476,24 @@ impl<'a, T: Read> Read for Buffer<'a, T> {
     fn read(&mut self, buf: &mut [u8]) -> io::Result<usize> {
         if self.buf_cur == self.buf_max {
             let len = self.reader.read(buf)?;
-            self.buf.extend_from_slice(&buf[..len]);
+            self.part.buf.extend_from_slice(&buf[..len]);
             Ok(len)
         } else {
-            let len = (&self.buf[self.buf_cur..self.buf_max]).read(buf)?;
+            let len = (&self.part.buf[self.buf_cur..self.buf_max]).read(buf)?;
             self.buf_cur += len;
             Ok(len)
         }
+    }
+}
+
+impl<'a, T> Buffer<'a, T> where T: std::io::Read {
+    // If we manage to read all the bytes, we reset the buffer
+    fn read_once(&mut self, buf: &mut [u8]) -> io::Result<usize> {
+        self.read_exact(buf)?;
+        let rlen = buf.len();
+        self.crc.update(buf);
+        self.part.buf.truncate(0);
+        return Ok(rlen);
     }
 }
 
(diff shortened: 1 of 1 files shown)


src/gz/bufread.rs
```diff
@@ -419,20 +568,20 @@ impl<R: BufRead> Read for GzDecoder<R> {
 
         loop {
             *inner = match mem::replace(inner, GzState::End) {
-                GzState::Header(mut buf) => {
+                GzState::Header(mut part) => {
                     let result = {
-                        let mut reader = Buffer::new(&mut buf, reader.get_mut().get_mut());
-                        read_gz_header(&mut reader)
+                        let mut reader = Buffer::new(&mut part, reader.get_mut().get_mut());
+                        read_gz_header_part(&mut reader)
```
catenacyber, line 574: Why is src/gz/write.rs using `read_gz_header` ?
I guess we should have only one function instead of both `read_gz_header` and `read_gz_header_part `

src/gz/bufread.rs
```diff
@@ -338,34 +476,45 @@ impl<'a, T: Read> Read for Buffer<'a, T> {
...
-            read_gz_header(&mut reader)
+            let mut reader = Buffer::new(&mut part, &mut r);
+            read_gz_header_part(&mut reader)
         };
 
         let state = match result {
-            Ok(hdr) => {
-                header = Some(hdr);
+            Ok(()) => {
+                header = Some(part.take_header());
```
catenacyber, line 514: `read_gz_header_part` cannot return directly the `GzHeader` structure because it would move part of the `GzHeaderPartial` which is behind a mutable reference.

Thus, the need to break in 2 phases : 
1 return `Ok`
2 take ownership of the partial header fields

src/gz/bufread.rs
```diff
@@ -338,34 +476,45 @@ impl<'a, T: Read> Read for Buffer<'a, T> {
...
+            let len = (&self.part.buf[self.buf_cur..self.buf_max]).read(buf)?;
             self.buf_cur += len;
             Ok(len)
         }
     }
 }
 
+impl<'a, T> Buffer<'a, T> where T: std::io::Read {
+    // If we manage to read all the bytes, we reset the buffer
+    fn read_once(&mut self, buf: &mut [u8]) -> io::Result<usize> {
```
catenacyber, line 491: I did not see any good fit in neither `BufRead` nor `Read` for this...

What I want is
`read_exact` but buffer it there is not enough data, and clear the buffer if it succeeded

Is there some more idiomatic rust way ?

src/gz/bufread.rs
```diff
@@ -40,6 +40,90 @@ fn read_le_u16<R: Read>(r: &mut R) -> io::Result<u16> {
...
+                if r.part.flg & FEXTRA != 0 {
+                    let mut extra = vec![0; r.part.xlen as usize];
+                    r.read_once(&mut extra)?;
+                    r.part.extra = Some(extra);
+                }
+                r.part.state = GzHeaderParsingState::GzHeaderParsingFilename;
+            }
+            GzHeaderParsingState::GzHeaderParsingFilename => {
+                if r.part.flg & FNAME != 0 {
+                    for byte in r.reader.bytes() {
```
catenacyber, line 89: I did not find any more idiomatic way to do this

src/gz/bufread.rs
```diff
@@ -338,34 +476,45 @@ impl<'a, T: Read> Read for Buffer<'a, T> {
...
         }
     }
 }
 
+impl<'a, T> Buffer<'a, T> where T: std::io::Read {
+    // If we manage to read all the bytes, we reset the buffer
+    fn read_once(&mut self, buf: &mut [u8]) -> io::Result<usize> {
+        self.read_exact(buf)?;
+        let rlen = buf.len();
+        self.crc.update(buf);
```
catenacyber, line 494: I did not find a way to chain with CrcReader because of this `read_once` logic

src/gz/bufread.rs
catenacyber: Maybe something like `r.part.filename = r.read_until_zero()?;` could be done ?

catenacyber:
I guess there should be some additions to the test framework as well. Like testing for some inputs that we get the same result if the reader is supplying bytes one at a time and we call multiple times `read` with treating the case `io::ErrorKind::WouldBlock` by making one more byte available...

src/gz/bufread.rs
```diff
@@ -304,9 +388,61 @@ pub struct GzDecoder<R> {
...
 }
 
+#[derive(Debug)]
+pub enum GzHeaderParsingState {
+    GzHeaderParsingStart = 0,
+    GzHeaderParsingXlen = 1,
+    GzHeaderParsingExtra = 2,
+    GzHeaderParsingFilename = 3,
+    GzHeaderParsingComment = 4,
+    GzHeaderParsingCrc = 5,
```
alexcrichton, line 398: I think the `= N` and the `GzHeaderParsing` prefix can both be dropped from each of these variants.

src/gz/bufread.rs
```diff
@@ -304,9 +388,61 @@ pub struct GzDecoder<R> {
...
+    buf: Vec<u8>,
+
+    state: GzHeaderParsingState,
+    flg: u8,
+    os: u8,
+    xlen: u16,
+    mtime: u32,
+    extra: Option<Vec<u8>>,
+    filename: Vec<u8>,
+    comment: Vec<u8>,
```
alexcrichton, line 412: Could this perhaps internally store a `GzHeader` and then it just appends a `buf` and `state` field?

src/gz/bufread.rs
```diff
@@ -40,6 +40,90 @@ fn read_le_u16<R: Read>(r: &mut R) -> io::Result<u16> {
...
+                        return Err(corrupt());
+                    }
+                }
+                return Ok(());
+            }
+        }
+    }
+}
+
 pub(crate) fn read_gz_header<R: Read>(r: &mut R) -> io::Result<GzHeader> {
```
alexcrichton, line 127: Shouldn't this function be replaced with the above one?

alexcrichton (commented):
Thanks!

Can you explain more about the `read_once` aspect? I'm not sure that the internal usage of `read_exact` will work well with async I/O where it may fail to make progress but may make progress later. It looks like in-progress bytes may be lost? 

Overall this looks pretty reasonable but I think the byte buffering could use some tweaks and/or documentation improvements.

src/gz/bufread.rs
catenacyber: Ok, will do

src/gz/bufread.rs
catenacyber: Good idea, will try

src/gz/bufread.rs
catenacyber: Indeed.
But it does not work for now as gzip writer uses `read_gz_header` with a different kind of reader
And I do not know why writer needs to call `read_gz_header`
cf https://github.com/rust-lang/flate2-rs/pull/278#discussion_r672140753

catenacyber:
> Can you explain more about the read_once aspect? I'm not sure that the internal usage of read_exact will work well with async I/O where it may fail to make progress but may make progress later. It looks like in-progress bytes may be lost?

It is possible I missed some points.
What I want `read_once` to do is to try to read the exact number of bytes of the buffer, and if there are not enough, save them into the structure's buffer...
Relooking at it, it looks like it does not do that yet... Will try something else

src/gz/bufread.rs
catenacyber: Ah, we need at least `flg` and `xlen` in `GzHeaderPartial`
So, should I still try to put a `GzHeader` in `GzHeaderPartial` ?

catenacyber:
I think `read_once` was correct because it is used with `Buffer` and not a generic `Read` which just keeps buffering the read data (until the buffer is reset/truncated in `read_once`)

src/gz/bufread.rs
alexcrichton: I don't know without doing it myself, but I'm trying to ideally reduce duplication here instead of listing the fields twice in a number of places.

src/gz/bufread.rs
alexcrichton: I don't know what that other call is doing, but I don't think there should be two functions to parse the header here.

alexcrichton:
I don't think `read_once` is working as intended. The `read_exact` method you're calling may call `read` multiple times and if it ends in an error then it discards all previously read data and doesn't return how much was read.

src/gz/bufread.rs
catenacyber: Right, doing so

src/gz/bufread.rs
catenacyber: Indeed,
Now `read_gz_header` just calls once `read_gz_header_part

catenacyber:
> I don't think read_once is working as intended.

You were right.

I added a test, and fixed `read_once` accordingly

src/gz/bufread.rs
```diff
@@ -639,3 +731,156 @@ impl<R: AsyncWrite + BufRead> AsyncWrite for MultiGzDecoder<R> {
...
+
+#[cfg(test)]
+mod tests {
+    use crate::gz::bufread::*;
+    use std::io;
+    use std::io::{Cursor, Read, Write};
+
+    //a cursor turning EOF into blocking errors
+    #[derive(Debug)]
+    pub struct BlockingCursor {
```
catenacyber, line 743: I do not know if there is such a structure ready in rust implementing `Read` and where we can specify at some points that we received more data

src/gz/bufread.rs
```diff
@@ -639,3 +731,156 @@ impl<R: AsyncWrite + BufRead> AsyncWrite for MultiGzDecoder<R> {
...
+        match r.write_all(&data2) {
+            Ok(()) => {}
+            _ => {
+                panic!("Unexpected result for write_all");
+            }
+        }
+        r.set_position(pos);
+
+        // Third read : still incomplete for 7 bytes (we have 5)
+        let mut reader2 = Buffer::new(&mut part, &mut r);
```
catenacyber, line 842: I need to redefine `reader`, otherwise I get
```
error[E0499]: cannot borrow `r` as mutable more than once at a time
   --> src/gz/bufread.rs:831:19
    |
810 |         let mut reader = Buffer::new(&mut part, &mut r);
    |                                                 ------ first mutable borrow occurs here
...
831 |         let pos = r.position();
    |                   ^ second mutable borrow occurs here
...
842 |         match reader.read_once(&mut out) {
    |               ------ first borrow later used here
```

I do not understand why `pos` a `u64` is considered a mutable borrow

src/gz/bufread.rs
```diff
@@ -34,87 +34,118 @@ fn bad_header() -> io::Error {
...
-
-    Ok(GzHeader {
-        extra: extra,
-        filename: filename,
-        comment: comment,
-        operating_system: os,
-        mtime: mtime,
-    })
+        Err(err) => {
+            return Err(err);
```
alexcrichton, line 146: Could this use `?` instead of explicit returns?

alexcrichton (commented):
I'm ok w/ whatever is necessary for tests, I don't mind too much about their idioms. I think `read_once` is still suffering the same partial-read bugs from before?

catenacyber:
> I think read_once is still suffering the same partial-read bugs from before?

The test should prove that `read_once` works at stage `Fourth read : now succesful for 7 bytes`
`read_once` calls `read_exact` which calls `read` multiple times, but each time the data gets buffered (for subsequent calls to `read_once` who did not finish)

src/gz/bufread.rs
catenacyber: indeed nicer

alexcrichton:
Hm ok there's some unusual stuff going on with `Buffer` type I don't remember at all, this will take a bit longer to page everything back in and try to understand...

catenacyber:
> Hm ok there's some unusual stuff going on with `Buffer` type I don't remember at all, this will take a bit longer to page everything back in and try to understand...

Yes `Buffer`'s `read` just saves every successful read bytes from the subsequent `Read` into an internal buffer
And in some case, it begins to return these saved bytes, before reading more from the `Read`

src/gz/bufread.rs
```diff
@@ -338,34 +401,50 @@ impl<'a, T: Read> Read for Buffer<'a, T> {
...
+    T: std::io::Read,
+{
+    // If we manage to read all the bytes, we reset the buffer
+    fn read_once(&mut self, buf: &mut [u8]) -> io::Result<usize> {
+        self.buf_cur = 0;
+        self.read_exact(buf)?;
+        let rlen = buf.len();
+        self.crc.update(buf);
+        self.part.buf.truncate(0);
+        self.buf_max = 0;
```
alexcrichton, line 425: Trying to read over this some more, I don't think I understand why this method needs to exist. I don't think that the manipulation of the fields of the structure here (e.g. the updates of `buf_{cur,max}` and the `truncate` are right. That seems like it's discarding data which might otherwise be succesfully read by subsequent calls to `read_once`. 

This seems like callers of this function should be calling `read_exact` and then on success updating the crc manually for the header? Other than that I'm not sure why this function is needed?

src/gz/bufread.rs
```diff
@@ -34,87 +34,107 @@ fn bad_header() -> io::Error {
...
+                r.part.state = GzHeaderParsingState::Filename;
+            }
+            GzHeaderParsingState::Filename => {
+                if r.part.flg & FNAME != 0 {
+                    if None == r.part.header.filename {
+                        let vec = Vec::new();
+                        r.part.header.filename = Some(vec);
+                    };
+                    if let Some(ref mut b) = r.part.header.filename {
+                        for byte in r.reader.bytes() {
```
alexcrichton, line 88: I don't think that reaching into the `reader` here is also quite right since it bypasses the buffering that's otherwise happening in `Buffer`?

catenacyber: I also think this is bad style even if it is intentional as the buffering takes directly place in the `GzHeader`
Should I rather add a method `read_until` to the `Buffer` ?

src/gz/bufread.rs
catenacyber: The problem with `read_exact` is that we do not reset the buffer, leading to the quadratic complexity as we keep parsing all over and again.

What I want is
`read_exact` but buffer it there is not enough data (that was already ok), and clear the buffer if it succeeded

For example, the reader first provides 7 bytes, `read_gz_header_part` buffers these 7 bytes and returns `ErrorKind::WouldBlock`

Then a second call to `read_gz_header_part`, the reader provides 4 more bytes
There is a successful read of the 10 bytes header
As we have flag `FEXTRA`, we try to read a `u16` but fail with `ErrorKind::WouldBlock`
The problem with the current code is that we have a 11 byte buffer instead of a 1 byte one for later calls to `read_gz_header_part`

By the way, I see a problem with my code that arc should belong to `GzHeaderPartial` as it lives longer than `Buffer`

alexcrichton: Nah yeah I understand the quadratic part, and I forgot that this is where the fix for that is going. To fix the quadratic bits, though, I think that the `.bytes()` iterator can be replaced when reading the nul-terminated string with custom logic to empty the buffer. Basically when reading the C strings for the various fields I think that we should consume the bytes and not buffer them in the general intermediate buffer, but rather buffer them in the location where the field is read from (basically store them directly where they're gonna go).

src/gz/bufread.rs
alexcrichton: I don't have a shape of code in mind, but there's a lot of confusing buffer-this-sometimes-dont logic going on here and there's very little documentation. I think something needs to be improved (code structure or similar) to make this more understandable.

catenacyber: I agree that this is not nice.
I just pushed something new.
The difficulty is `r.bytes()` is a mutable borrow and we want another mutable borrow for `r.part.header.filename`
I did not manage to have a mutable reference to `r.part.header.filename` in `Buffer` because of lifetime issues such as 
```
error[E0623]: lifetime mismatch
  --> src/gz/bufread.rs:89:39
   |
46 | fn read_gz_header_part<R: Read>(r: &mut Buffer<R>) -> io::Result<()> {
   |                                    --------------
   |                                    |
   |                                    these two types are declared with different lifetimes...
...
89 |                     r.direct_buffer = r.part.header.filename.as_ref();
   |                                       ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^ ...but data from `r` flows into `r` here
```

So, in the end `Buffer::read` looks at the `GzHeaderParsingState` to decide where to buffer the data

Another (not so nice) pattern could be
```
                    r.direct = true;
                    let mut bytebuf = [0; 1];
                    loop {
                        r.read_exact(&mut bytebuf)?;
                        if bytebuf[0] == 0 {
                            r.direct = false;
                            break;
                        }
                        if let Some(ref mut b) = r.part.header.filename {
                            b.push(bytebuf[0]);
                        }
                    }
```

src/gz/bufread.rs
catenacyber: >  Basically when reading the C strings for the various fields I think that we should consume the bytes and not buffer them in the general intermediate buffer, but rather buffer them in the location where the field is read from (basically store them directly where they're gonna go).

I did that, so `Buffer::read` looks at GzHeaderParsingState to decide where to buffer the data.

The point I do not like about this is that the logic is now split between `read_gz_header_part` and `Buffer::read`, ie we do not see in `read_gz_header_part` that we are filling the filename

catenacyber:
Thanks @alexcrichton 
When would this be available in a release ?

alexcrichton:
I'll go ahead and release now

catenacyber:
Thanks Alex

PR #279, opened 2021-07-29
Convert bad file to valid hex file

jose-pr:
Fixes what https://github.com/rust-lang/flate2-rs/pull/277 tried to fix, changing the name alone wont fix the problem as the renamed file still contains the GZ magic number and will still be detected by a file scanner as a corrupt GZ file.
Used the following to generate the hex file.

```rust
fn bin_to_hex(bin_reader: impl Read, hex_writer: impl Write) {
    let bin_reader = BufReader::new( bin_reader);
    let mut hex_writer = BufWriter::new( hex_writer);

    for byte in bin_reader.bytes() {
        let byte = byte.unwrap();
        write!(hex_writer, "{:X} ", byte).unwrap();
    }

    hex_writer.flush().unwrap();
}
```


alexcrichton:
Sorry I don't want to make files like this too complicated in this repo, and I think that this is a step too far. I think if automated scanners are tripping up over this file then there's only so much this repository can do before something should be fixed in the scanner itself.

PR #280, opened 2021-09-13
gzdecoder: fix endless loop in read implementation

catenacyber:
Introduced by commit 7212da84fa0a9fe77c240e95a675f0567e9d698c
Adds a test to prevent this behavior

@alexcrichton sorry for turning a quadratic complexity into an endless loop 

You can check that the test does not end without the patch in `impl<R: BufRead> Read for GzDecoder<R>`


alexcrichton:
No worries, thanks!

PR #283, opened 2021-12-02
Make Clippy happier, minor cleanups

nyurik:
I ran `cargo clippy` and addressed some issues it raised, while leaving other issues for another time. Hopefully this will make this code a bit easier for newcomers to read.

Biggest changes:
* `foo.len() > 0` ➡ `!foo.is_empty()`
* `return x;` ➡ `x`
* `{ foo: foo }` ➡ `{ foo }`
* adding a few `Default` trait implementations
* some `match` statement cleanup

Also, fixed two `ffi::uInt::max_value()` ➡ `ffi::uInt::MAX` due to deprecation.

Biggest TODOs (please give feedback if they should be fixed too):
* Clippy has complained about many `foo >> 0` -- which is a noop that might confuse (?)
* some other corner cases that I don't know enough about

### DIFF
--- a/src/bufreader.rs
+++ b/src/bufreader.rs
@@ -42,7 +42,7 @@ impl<R: Read> BufReader<R> {
 
     pub fn with_buf(buf: Vec<u8>, inner: R) -> BufReader<R> {
         BufReader {
-            inner: inner,
+            inner,
             buf: buf.into_boxed_slice(),
             pos: 0,
             cap: 0,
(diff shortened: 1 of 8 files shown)


nyurik:
@alexcrichton thx for merging! Do you have any thoughts about the TODO in the comment?

alexcrichton:
Sorry no, I can perhaps offer my thoughts on a case-by-case basis but I don't run clippy myself.

Issue #274, opened 2021-06-04
Don't depend on libc crate when only compiled with rust_backend


jrmuizel:
It shouldn't be need in that case.

Craig-Macomber:
It looks like libc is needed (in the top level Cargo.toml) only if `cfg(any(feature = "miniz-sys", feature = "any_zlib"))`.

My understanding of the Cargo.toml format is we don't have a way to deal with the "miniz-sys" case without making a breaking change (or some other significant refactoring, like adding an indirection crate so we avoid having a direct dep named "miniz-sys"), since users of this crate using that feature, but not the default feature, have no way to get opted into an getting libc, since we can't have a feature name that overlaps an optional dependency, and can't make the optional dependency cause another dependency.

Here is a fix, though as noted above, this is a breaking change for users of the miniz-sys feature: https://github.com/Craig-Macomber/flate2-rs/tree/un-libc

oyvindln:
If the only parts of libc used are the C data types, it's possible you could do a workaround using a cut-down version of libc or similar (not sure if there are any existing crates for this) with only that included. There is something [similar](https://github.com/Frommi/miniz_oxide/blob/cdecae0531f21099183375c62bdcf30d0292f330/src/lib.rs#L17) in the C API part of miniz_oxide though it's not really used for much atm, though one would need a more extensive one to account for different platforms probably. Maybe a case for a new libc_types crate or similar.

LingMan:
The libc dependency was recently eliminated entirely (https://github.com/rust-lang/flate2-rs/commit/a4628b7955a5564e1fb8ce37ed18affb09f42b18), so I guess this issue could be closed?

JohnTitor:
Closing as fixed

PR #290, opened 2022-01-18
remove libc from default features

Craig-Macomber:
This also required adding an alias for the miniz-sys feature which includes libc, and this is a breaking change for users of the miniz-sys feature: they need to use `miniz` instead or add the `libc` feature.

I am unaware of a way to make this a non-breaking change, and still remove libc from the default.

Another option would be to include libc in the default feature, but not in an explicit rust backend. That would avoid breaking users of miniz-sys that include the default feature, but it would still break users who use it without the default.

This addresses issue #274

### DIFF
--- a/.github/workflows/main.yml
+++ b/.github/workflows/main.yml
@@ -29,39 +29,39 @@ jobs:
             os: windows-latest
             rust: stable-x86_64-gnu
     steps:
-      - uses: actions/checkout@master
-      - name: Install Rust (rustup)
-        run: rustup update ${{ matrix.rust }} --no-self-update && rustup default ${{ matrix.rust }}
-        shell: bash
-      - run: cargo build
-      - run: rustdoc --test README.md -L target/debug/deps --extern flate2=target/debug/libflate2.rlib --edition=2018
-      - run: cargo test
-      - run: cargo test --features zlib
-      - run: cargo test --features miniz
-      - run: cargo test --features zlib --no-default-features
-      - run: cargo test --features zlib-ng-compat --no-default-features
-      - run: cargo test --features cloudflare_zlib --no-default-features
-        if: matrix.build != 'mingw'
-      - run: cargo test --features miniz --no-default-features
-      - run: cargo test --features tokio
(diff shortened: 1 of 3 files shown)


Craig-Macomber:
General disclaimer: I'm very new to optional features, so I may have missed something. If you do want to accept this change, despite its breaking nature, don't assume it did it correctly. I think I ran the tests correctly with the default rust back-end and miniz, but thats all I tried. Also I'm not really sure what the proper approach for testing combinations of features is, or what is acceptable breaking change wise (ex: does this change need to be delayed until we want to do do a new major version?)

Craig-Macomber:
Looking a little closer, miniz-sys is published within this repo, so renaming it might not be too hard. That would allow using the feature name miniz-sys like we used to to avoid the change being breaking. If this approach would be preferred, recommend a name and I'll do the change.

alexcrichton:
I believe cargo's very recently stabilized namespaced features feature will enable solving this, but it would probably be best to wait for that to avoid any breakage

Craig-Macomber:
> I believe cargo's very recently stabilized namespaced features feature will enable solving this, but it would probably be best to wait for that to avoid any breakage

I think you are referring to: https://doc.rust-lang.org/cargo/reference/unstable.html#namespaced-features
Which looks like it was stabilized in https://github.com/rust-lang/cargo/pull/10269

I'll update this PR to use it, then you can merge it whenever you think its been long enough that its ok to depend on that cargo feature.

Looks like this will be waiting for us to be ok to require Cargo 1.60, which isn't even scheduled for release until 2022-04-07 (https://github.com/rust-lang/cargo/blob/master/CHANGELOG.md#cargo-160-2022-04-07) so it will be a while.

When the ecosystem is ready to use that, it will be a great time to clean up lots of package's dependencies :)

Craig-Macomber:
I created https://github.com/rust-lang/flate2-rs/pull/291 with the new approach (it requires much less changes than on this branch, so I made it separately), which I left as a draft since it should not be merged for several months waiting for a new cargo release at least.

PR #291, opened 2022-01-19
make libc optional

Craig-Macomber:
Use the new namespaced-features cargo feature to make libc optional, and not included by default without breaking users of the miniz-sys feature.

This currently requires cargo nightly with `-Z namespaced-features`, and should be in stable in the [1.60 release, scheduled for 2022-04-07](https://github.com/rust-lang/cargo/blob/master/CHANGELOG.md#cargo-160-2022-04-07). Sometime after that, this should be safe to merge.

This PR replaces https://github.com/rust-lang/flate2-rs/pull/290 which did the same thing, but as a breaking change.

This fixes https://github.com/rust-lang/flate2-rs/issues/274


alexcrichton:
Awesome, thanks!

Craig-Macomber:
I'm not sure when/if you want to bump the minimum supported rust version to 1.60, but with 1.60 stable now, it is possible to merge this if you want.

joshtriplett:
@Craig-Macomber https://github.com/rust-lang/flate2-rs/pull/303 should make this easier, and avoid the need for the `dep:` syntax.

joshtriplett:
@Craig-Macomber I've now pushed a change that eliminates the use of libc entirely; with miniz-sys support gone, we no longer need it at all.

PR #292, opened 2022-01-19
Remove `tokio` support

Expyron:
Following discussion in #214

The current support is for tokio 0.1, which is very outdated.

Crate `async-compression` provides async integration if needed, so rather than upgrading to the latest version of tokio, this is a blanket removal of the feature.

This removal is a breaking change only if the feature is enabled, so I'm not sure how it should be tied to a version change.
I cannot find any dependent crate on crates.io which uses the `tokio` feature


alexcrichton:
Seems reasonable to me, thanks for the PR!

Issue #178, opened 2018-12-05
Add example & references in docs to steer people directly to MultiGzDecoder


mdekstrand:
I think it would help users to have documentation that clearly points people to `MultiGzDecoder` for decompressing gzipped data. I see two specific places for this:

1. On the front page, in addition to the encoding example, give a decoding example.
2. In the `GzDecoder` docs, make a clear warning that `GzDecoder` only decodes a single stream, and if users want to reliably decode arbitrary gzipped files, they should look at `MultiGzDecoder`.

'I have some gzipped data, please decode' seems like a very common use case for `flate2` (it was certainly my use case). Right now, it isn't obvious at a quick intro level how to do that.

alexcrichton:
Sounds plausible to me!

aeshirey:
I've been burned by using `GzDecoder`, not knowing there were multiple members in the file. Even if you know an input is a single member, is there a compelling reason to use `GzDecoder` over `MultiGzDecoder`? (Presumably the latter will yield the same results, but is there a perf hit for it?)

mdekstrand:
There are some obscure cases, but more around when you know that a file is *multiple* streams. The Alpine `.apk` format uses multiple `.gz` streams in a design that is simultaneously clever and mildly horrifying. They tar & compress the package content; then tar & compress metadata, including a checksum of the compressed content stream, as a “tar fragment” (sequence of tar entries without the end-of-tarball indicator) and prepend it to the content; and finally, they sign all of *that*, pack the signature into a tar entry, gzip it, and prepend it to the whole thing. So you can `tar xvzf` the entire file and get the content, metadata, and signature, but if you want to work with the file directly and verify its signature, you need to know when the first gzip stream stops.

I don't think doing surgery on Alpine packages is the common use case for `flate2`.

Issue #192, opened 2019-04-25
Reading multiple GZIP archives in one file, individually


matthewrobertbell:
Hi,

I have a large .gz file with many (hundreds of thousands) individual archives in the one file.

Using GZDecoder works on this, for the first archive. Using MultiGZDecoder works, allowing me to produce one big string with data from all archives.

Since the archives are to be processed individually, and because it is hard to split the parts of data once they are all in one big string, I would like to produce something like an iterator of strings, one per archive. Is this possible currently? If not, is it something that is easy to build on top of flate2?

Thanks!

alexcrichton:
I think you'd probably want to see how `MultiGzDecoder` works internally and probably copy that externally. I don't think it would be too too hard to set up though!

PR #324, opened 2022-10-31
Recommend MultiGzDecoder over GzDecoder in docs

jsha:
Part of #178.

I removed the text that says "the specification, however, allows ..." and the comment about bioinformatics because they make it sound like MultiGzDecoder is a rare thing that you should only enable if you need it, but it's actually the correct choice for almost all cases since it implements what the RFC considers a "gzip file."

### DIFF
--- a/src/gz/bufread.rs
+++ b/src/gz/bufread.rs
@@ -172,8 +172,8 @@ impl<R: BufRead + Write> Write for GzEncoder<R> {
 /// This structure exposes a [`BufRead`] interface, reading compressed data
 /// from the underlying reader, and emitting uncompressed data.
 ///
-/// After reading a single member of the gzip data this reader will return 
-/// Ok(0) even if there are more bytes available in the underlying reader. 
+/// After reading a single member of the gzip data this reader will return
+/// Ok(0) even if there are more bytes available in the underlying reader.
 /// If you need the following bytes, call `into_inner()` after Ok(0) to
 /// recover the underlying reader.
 ///
--- a/src/gz/bufread.rs
+++ b/src/gz/bufread.rs
@@ -167,8 +167,7 @@ impl<R: BufRead + Write> Write for GzEncoder<R> {
     }
 }
 
-/// A decoder for a single member of a gzip file. Prefer [MultiGzDecoder] for
-/// most uses.
+/// A decoder for a gzip file with a single member.
 ///
 /// This structure consumes a [`BufRead`] interface, reading compressed data
 /// from the underlying reader, and emitting uncompressed data.
@@ -397,11 +396,12 @@ impl<R: BufRead + Write> Write for GzDecoder<R> {
     }
 }
 
-/// A gzip streaming decoder that decodes a complete [gzip file].
+/// A gzip streaming decoder that decodes a [gzip file] with multiple members.
 ///
 /// A gzip file consists of a series of "members" concatenated one after another.
 /// MultiGzDecoder decodes all members of a file, while [GzDecoder] will only decode
-/// the first member. MultiGzDecoder is preferable in most cases.
+/// the first member. Learn more
+/// [in the introduction](https://docs.rs/flate2/*/flate2/#About-multi-member-Gzip-files).
 ///
 /// This structure exposes a [`BufRead`] interface that will consume all gzip members
 /// from the underlying reader and emit uncompressed data.
--- a/src/gz/bufread.rs
+++ b/src/gz/bufread.rs
@@ -167,11 +167,22 @@ impl<R: BufRead + Write> Write for GzEncoder<R> {
     }
 }
 
-/// A decoder for a gzip file with a single member.
+/// A decoder for the first member of a [gzip file].
 ///
-/// This structure consumes a [`BufRead`] interface, reading compressed data
+/// This structure exposes a [`BufRead`] interface, reading compressed data
 /// from the underlying reader, and emitting uncompressed data.
 ///
+/// After reading the first member of a gzip file (which is often, but not
+/// always, the only member), this reader will return Ok(0) even if there
+/// are more bytes available in the underlying reader. If you want to be sure
+/// not to drop bytes on the floor, call `into_inner()` after Ok(0) to
+/// recover the underlying reader.
+///
+/// To handle gzip files that may have multiple members, see [`MultiGzDecoder`]
+/// or read more
+/// [in the introduction](../index.html#about-multi-member-gzip-files).
+///
+/// [gzip file]: https://www.rfc-editor.org/rfc/rfc1952#page-5
 /// [`BufRead`]: https://doc.rust-lang.org/std/io/trait.BufRead.html
 ///
 /// # Examples
@@ -396,15 +407,17 @@ impl<R: BufRead + Write> Write for GzDecoder<R> {
     }
 }
 
-/// A gzip streaming decoder that decodes a [gzip file] with multiple members.
+/// A gzip streaming decoder that decodes a [gzip file] that may have multiple members.
 ///
-/// A gzip file consists of a series of "members" concatenated one after another.
-/// MultiGzDecoder decodes all members of a file, while [GzDecoder] will only decode
-/// the first member. Learn more
-/// [in the introduction](https://docs.rs/flate2/*/flate2/#About-multi-member-Gzip-files).
+/// This structure exposes a [`BufRead`] interface that will consume compressed
+/// data from the underlying reader and emit uncompressed data.
 ///
-/// This structure exposes a [`BufRead`] interface that will consume all gzip members
-/// from the underlying reader and emit uncompressed data.
+/// A gzip file consists of a series of *members* concatenated one after another.
+/// MultiGzDecoder decodes all members of a file and returns Ok(0) once the
+/// underlying reader does.
+///
+/// To handle members seperately, see [GzDecoder] or read more
+/// [in the introduction](../index.html#about-multi-member-gzip-files).
 ///
 /// [gzip file]: https://www.rfc-editor.org/rfc/rfc1952#page-5
 /// [`BufRead`]: https://doc.rust-lang.org/std/io/trait.BufRead.html
--- a/src/gz/bufread.rs
+++ b/src/gz/bufread.rs
@@ -167,11 +167,11 @@ impl<R: BufRead + Write> Write for GzEncoder<R> {
     }
 }
 
-/// A gzip streaming decoder
+/// A decoder for a single member of a gzip file. Prefer [MultiGzDecoder] for
+/// most uses.
 ///
 /// This structure consumes a [`BufRead`] interface, reading compressed data
 /// from the underlying reader, and emitting uncompressed data.
-/// Use [`MultiGzDecoder`] if your file has multiple streams.
 ///
 /// [`BufRead`]: https://doc.rust-lang.org/std/io/trait.BufRead.html
 ///
@@ -397,20 +397,16 @@ impl<R: BufRead + Write> Write for GzDecoder<R> {
     }
 }
 
-/// A gzip streaming decoder that decodes all members of a multistream
+/// A gzip streaming decoder that decodes a full [gzip file].
 ///
-/// A gzip member consists of a header, compressed data and a trailer. The [gzip
-/// specification](https://tools.ietf.org/html/rfc1952), however, allows multiple
-/// gzip members to be joined in a single stream. `MultiGzDecoder` will
-/// decode all consecutive members while [`GzDecoder`] will only decompress
-/// the first gzip member. The multistream format is commonly used in
-/// bioinformatics, for example when using the BGZF compressed data. It's also useful
-/// to compress large amounts of data in parallel where each thread produces one stream
-/// for a chunk of input data.
+/// A gzip file consists of a series of "members" concatenated one after another.
+/// MultiGzDecoder decodes all members of a file, while [GzDecoder] will only decode
+/// the first one member. MultiGzDecoder is preferable in most cases.
 ///
 /// This structure exposes a [`BufRead`] interface that will consume all gzip members
 /// from the underlying reader and emit uncompressed data.
 ///
+/// [gzip file]: https://www.rfc-editor.org/rfc/rfc1952#page-5
 /// [`BufRead`]: https://doc.rust-lang.org/std/io/trait.BufRead.html
 ///
 /// # Examples
--- a/src/gz/bufread.rs
+++ b/src/gz/bufread.rs
@@ -167,15 +167,14 @@ impl<R: BufRead + Write> Write for GzEncoder<R> {
     }
 }
 
-/// A decoder for the first member of a [gzip file].
+/// A decoder for a single member of a [gzip file].
 ///
 /// This structure exposes a [`BufRead`] interface, reading compressed data
 /// from the underlying reader, and emitting uncompressed data.
 ///
-/// After reading the first member of a gzip file (which is often, but not
-/// always, the only member), this reader will return Ok(0) even if there
-/// are more bytes available in the underlying reader. If you want to be sure
-/// not to drop bytes on the floor, call `into_inner()` after Ok(0) to
+/// After reading a single member of the gzip data this reader will return 
+/// Ok(0) even if there are more bytes available in the underlying reader. 
+/// If you need the following bytes, call `into_inner()` after Ok(0) to
 /// recover the underlying reader.
 ///
 /// To handle gzip files that may have multiple members, see [`MultiGzDecoder`]
@@ -413,8 +412,8 @@ impl<R: BufRead + Write> Write for GzDecoder<R> {
 /// data from the underlying reader and emit uncompressed data.
 ///
 /// A gzip file consists of a series of *members* concatenated one after another.
-/// MultiGzDecoder decodes all members of a file and returns Ok(0) once the
-/// underlying reader does.
+/// MultiGzDecoder decodes all members from the data and only returns Ok(0) when the
+/// underlying reader does. For a file, this reads to the end of the file.
 ///
 /// To handle members seperately, see [GzDecoder] or read more
 /// [in the introduction](../index.html#about-multi-member-gzip-files).
--- a/src/gz/bufread.rs
+++ b/src/gz/bufread.rs
@@ -397,11 +397,11 @@ impl<R: BufRead + Write> Write for GzDecoder<R> {
     }
 }
 
-/// A gzip streaming decoder that decodes a full [gzip file].
+/// A gzip streaming decoder that decodes a complete [gzip file].
 ///
 /// A gzip file consists of a series of "members" concatenated one after another.
 /// MultiGzDecoder decodes all members of a file, while [GzDecoder] will only decode
-/// the first one member. MultiGzDecoder is preferable in most cases.
+/// the first member. MultiGzDecoder is preferable in most cases.
 ///
 /// This structure exposes a [`BufRead`] interface that will consume all gzip members
 /// from the underlying reader and emit uncompressed data.
--- a/src/gz/read.rs
+++ b/src/gz/read.rs
@@ -95,11 +95,11 @@ impl<R: Read + Write> Write for GzEncoder<R> {
 /// This structure exposes a [`Read`] interface that will consume compressed
 /// data from the underlying reader and emit uncompressed data.
 ///
-/// After reading the first member of a gzip file (which is often, but not
-/// always, the only member), this reader will return Ok(0) even if there
-/// are more bytes available in the underlying reader. If you want to be sure
-/// not to drop bytes on the floor, call `into_inner()` after Ok(0) to
-/// recover the underlying reader.
+/// After reading a single member of the gzip data this reader will return
+/// Ok(0) even if there are more bytes available in the underlying reader.
+/// `GzDecoder` may have read additional bytes past the end of the gzip data.
+/// If you need the following bytes, wrap the `Reader` in a `std::io::BufReader` 
+/// and use `bufread::GzDecoder` instead.
 ///
 /// To handle gzip files that may have multiple members, see [`MultiGzDecoder`]
 /// or read more
--- a/src/gz/read.rs
+++ b/src/gz/read.rs
@@ -90,13 +90,22 @@ impl<R: Read + Write> Write for GzEncoder<R> {
     }
 }
 
-/// A decoder for a gzip file with a single member.
+/// A decoder for the first member of a [gzip file].
 ///
 /// This structure exposes a [`Read`] interface that will consume compressed
 /// data from the underlying reader and emit uncompressed data.
 ///
-/// This decoder only handles gzipped data with a single stream.
-/// Use [`MultiGzDecoder`] for gzipped data with multiple streams.
+/// After reading the first member of a gzip file (which is often, but not
+/// always, the only member), this reader will return Ok(0) even if there
+/// are more bytes available in the underlying reader. If you want to be sure
+/// not to drop bytes on the floor, call `into_inner()` after Ok(0) to
+/// recover the underlying reader.
+///
+/// To handle gzip files that may have multiple members, see [`MultiGzDecoder`]
+/// or read more
+/// [in the introduction](../index.html#about-multi-member-gzip-files).
+///
+/// [gzip file]: https://www.rfc-editor.org/rfc/rfc1952#page-5
 ///
 /// # Examples
 ///
@@ -180,15 +189,17 @@ impl<R: Read + Write> Write for GzDecoder<R> {
     }
 }
 
-/// A gzip streaming decoder that decodes a [gzip file] with multiple members.
+/// A gzip streaming decoder that decodes a [gzip file] that may have multiple members.
 ///
-/// A gzip file consists of a series of "members" concatenated one after another.
-/// MultiGzDecoder decodes all members of a file, while [GzDecoder] will only decode
-/// the first member. Learn more
-/// [in the introduction](https://docs.rs/flate2/*/flate2/#About-multi-member-Gzip-files).
+/// This structure exposes a [`Read`] interface that will consume compressed
+/// data from the underlying reader and emit uncompressed data.
 ///
-/// This structure exposes a [`Read`] interface that will consume all gzip members
-/// from the underlying reader and emit uncompressed data.
+/// A gzip file consists of a series of *members* concatenated one after another.
+/// MultiGzDecoder decodes all members of a file and returns Ok(0) once the
+/// underlying reader does.
+///
+/// To handle members seperately, see [GzDecoder] or read more
+/// [in the introduction](../index.html#about-multi-member-gzip-files).
 ///
 /// [gzip file]: https://www.rfc-editor.org/rfc/rfc1952#page-5
 ///
--- a/src/gz/read.rs
+++ b/src/gz/read.rs
@@ -90,11 +90,13 @@ impl<R: Read + Write> Write for GzEncoder<R> {
     }
 }
 
-/// A decoder for a single member of a gzip file. Prefer [MultiGzDecoder] for
-/// most uses.
+/// A decoder for a gzip file with a single member.
 ///
 /// This structure exposes a [`Read`] interface that will consume compressed
 /// data from the underlying reader and emit uncompressed data.
+///
+/// This decoder only handles gzipped data with a single stream.
+/// Use [`MultiGzDecoder`] for gzipped data with multiple streams.
 ///
 /// # Examples
 ///
@@ -178,11 +180,12 @@ impl<R: Read + Write> Write for GzDecoder<R> {
     }
 }
 
-/// A gzip streaming decoder that decodes a full [gzip file].
+/// A gzip streaming decoder that decodes a [gzip file] with multiple members.
 ///
 /// A gzip file consists of a series of "members" concatenated one after another.
 /// MultiGzDecoder decodes all members of a file, while [GzDecoder] will only decode
-/// the first member. MultiGzDecoder is preferable in most cases.
+/// the first member. Learn more
+/// [in the introduction](https://docs.rs/flate2/*/flate2/#About-multi-member-Gzip-files).
 ///
 /// This structure exposes a [`Read`] interface that will consume all gzip members
 /// from the underlying reader and emit uncompressed data.
--- a/src/gz/read.rs
+++ b/src/gz/read.rs
@@ -90,13 +90,11 @@ impl<R: Read + Write> Write for GzEncoder<R> {
     }
 }
 
-/// A gzip streaming decoder
+/// A decoder for a single member of a gzip file. Prefer [MultiGzDecoder] for
+/// most uses.
 ///
 /// This structure exposes a [`Read`] interface that will consume compressed
 /// data from the underlying reader and emit uncompressed data.
-/// Use [`MultiGzDecoder`] if your file has multiple streams.
-///
-/// [`Read`]: https://doc.rust-lang.org/std/io/trait.Read.html
 ///
 /// # Examples
 ///
@@ -180,21 +178,16 @@ impl<R: Read + Write> Write for GzDecoder<R> {
     }
 }
 
-/// A gzip streaming decoder that decodes all members of a multistream
+/// A gzip streaming decoder that decodes a full [gzip file].
 ///
-/// A gzip member consists of a header, compressed data and a trailer. The [gzip
-/// specification](https://tools.ietf.org/html/rfc1952), however, allows multiple
-/// gzip members to be joined in a single stream.  `MultiGzDecoder` will
-/// decode all consecutive members while [`GzDecoder`] will only decompress the
-/// first gzip member. The multistream format is commonly used in bioinformatics,
-/// for example when using the BGZF compressed data. It's also useful
-/// to compress large amounts of data in parallel where each thread produces one stream
-/// for a chunk of input data.
+/// A gzip file consists of a series of "members" concatenated one after another.
+/// MultiGzDecoder decodes all members of a file, while [GzDecoder] will only decode
+/// the first one member. MultiGzDecoder is preferable in most cases.
 ///
 /// This structure exposes a [`Read`] interface that will consume all gzip members
 /// from the underlying reader and emit uncompressed data.
 ///
-/// [`Read`]: https://doc.rust-lang.org/std/io/trait.Read.html
+/// [gzip file]: https://www.rfc-editor.org/rfc/rfc1952#page-5
 ///
 /// # Examples
 ///
--- a/src/gz/read.rs
+++ b/src/gz/read.rs
@@ -98,7 +98,7 @@ impl<R: Read + Write> Write for GzEncoder<R> {
 /// After reading a single member of the gzip data this reader will return
 /// Ok(0) even if there are more bytes available in the underlying reader.
 /// `GzDecoder` may have read additional bytes past the end of the gzip data.
-/// If you need the following bytes, wrap the `Reader` in a `std::io::BufReader` 
+/// If you need the following bytes, wrap the `Reader` in a `std::io::BufReader`
 /// and use `bufread::GzDecoder` instead.
 ///
 /// To handle gzip files that may have multiple members, see [`MultiGzDecoder`]
--- a/src/gz/read.rs
+++ b/src/gz/read.rs
@@ -182,7 +182,7 @@ impl<R: Read + Write> Write for GzDecoder<R> {
 ///
 /// A gzip file consists of a series of "members" concatenated one after another.
 /// MultiGzDecoder decodes all members of a file, while [GzDecoder] will only decode
-/// the first one member. MultiGzDecoder is preferable in most cases.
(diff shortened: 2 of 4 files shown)


src/gz/bufread.rs
```diff
@@ -270,7 +270,8 @@ impl<R: BufRead + Write> Write for GzEncoder<R> {
     }
 }
 
-/// A gzip streaming decoder
+/// Decode a single member of a gzip file. Prefer [MultiGzDecoder] for
+/// most uses.
```
JohnTitor, line 274: Since the struct itself does NOT decode, an expression like "A decoder..." is appropriate.

jsha: Fixed!

jsha:
I see you've also made some tweaks to these docs in https://github.com/rust-lang/flate2-rs/pull/347. Thanks! I went ahead and updated this PR, and I still think it's worthwhile.

In this PR I reference the concept of the "gzip file" instead of multistream, which is what the RFC talks about as the core concept (with "members" being a sub-concept).

Also, I removed the list of cases where gzip files containing multiple members are commonly used, since that seems to suggest that gzip files containing multiple members are only used in particular endeavors, when really they can be used any time you have a .gz file or a `Content-Encoding: gzip` HTTP response.

Similarly, I removed the comment "Use [`MultiGzDecoder`] if your file has multiple streams" because in many cases the user of this library doesn't know in advance if they will be processing a file with multiple streams or not; but it is always correct to use a MultiGzDecoder because it will handle a single-stream file just fine.

Byron (commented):
Thanks a lot! 

I feel the narrative is now inverted, and the way it should be, by making `MultiGzDecoder` effectively the default. To me it feels like ultimately, `MutliGzDecoder` really wants to be the default also by name, as right now it still feels special in that regard. Of course, I am not proposing to make name changes to types in this PR, which already gets the documentation to a place where most people who want to use `GzDecoder` will end up using the `MutiGzDecoder` instead.

src/gz/bufread.rs
```diff
@@ -397,20 +397,16 @@ impl<R: BufRead + Write> Write for GzDecoder<R> {
...
-/// specification](https://tools.ietf.org/html/rfc1952), however, allows multiple
-/// gzip members to be joined in a single stream. `MultiGzDecoder` will
-/// decode all consecutive members while [`GzDecoder`] will only decompress
-/// the first gzip member. The multistream format is commonly used in
-/// bioinformatics, for example when using the BGZF compressed data. It's also useful
-/// to compress large amounts of data in parallel where each thread produces one stream
-/// for a chunk of input data.
+/// A gzip file consists of a series of "members" concatenated one after another.
+/// MultiGzDecoder decodes all members of a file, while [GzDecoder] will only decode
+/// the first one member. MultiGzDecoder is preferable in most cases.
```
jongiddy, line 404: This should be either `the first member` or `one member`. `the first` is more informative.

Byron: I have applied this change.

Byron:
I think it would be nice if @JohnTitor could have another look. Overall, I believe this PR would help to get a better understanding on the intended use of `MultiGzDecoder`, and I would definitely have benefitted from that while working with #348 . So I see a lot of value in getting it merged. Thank you.

src/gz/bufread.rs
jongiddy: The same text appears in the `read` and `write` modules too.

Byron: Thanks for the catch, I applied the fix to the remaining copies of the text.

Byron (approved):

joshtriplett:
Whether the specification supports it or not, *in many cases* I think people will be surprised if they encounter a gzip stream with multiple files in it. I think that people typically think of gzip as a single-file format, and combine it with things like tar when storing multiple files.

So, I *don't* think we should go as far as saying that most people should use MultiGzDecoder. I think it's perfectly fine to adjust the documentation to say that it's allowed, and explain the distinction, but I think we should actually *explain* the distinction and the common usage rather than presenting MultiGzDecoder as the most common case.

src/gz/write.rs
```diff
@@ -373,17 +373,16 @@ impl<W: Read + Write> Read for GzDecoder<W> {
...
 ///
-/// A gzip member consists of a header, compressed data and a trailer. The [gzip
-/// specification](https://tools.ietf.org/html/rfc1952), however, allows multiple
-/// gzip members to be joined in a single stream. `MultiGzDecoder` will
-/// decode all consecutive members while `GzDecoder` will only decompress
-/// the first gzip member. The multistream format is commonly used in
-/// bioinformatics, for example when using the BGZF compressed data.
+/// A gzip file consists of a series of "members" concatenated one after another.
+/// MultiGzDecoder decodes all members of a file, while [GzDecoder] will only decode
+/// the first member. MultiGzDecoder is preferable in most cases.
```
joshtriplett, line 380:
```suggestion
/// the first member. Many gzip files in the wild will typically have one member,
/// but not all. Use MultiGzDecoder if you need to handle multiple members, or
/// [GzDecoder] if you know you only need to handle a single member.
```

src/gz/write.rs
```diff
@@ -373,17 +373,16 @@ impl<W: Read + Write> Read for GzDecoder<W> {
     }
 }
 
-/// A gzip streaming decoder that decodes all members of a multistream
+/// A gzip streaming decoder that decodes a full [gzip file].
```
joshtriplett, line 376:
```suggestion
/// A gzip streaming decoder that decodes a [gzip file] with multiple members.
```

src/gz/write.rs
```diff
@@ -166,11 +166,11 @@ impl<W: Write> Drop for GzEncoder<W> {
     }
 }
 
-/// A gzip streaming decoder
+/// A decoder for a single member of a gzip file. Prefer [MultiGzDecoder] for
+/// most uses.
```
joshtriplett, line 170:
```suggestion
/// A decoder for a gzip file with a single member.
```

I'd agree that we shouldn't *just* present this as a decoder for gzip in general, and should capture the distinction in the summary, but I'd stop short of telling people to *prefer* MultiGzDecoder.

src/gz/write.rs
```diff
@@ -166,11 +166,11 @@ impl<W: Write> Drop for GzEncoder<W> {
...
 ///
 /// This structure exposes a [`Write`] interface that will emit uncompressed data
 /// to the underlying writer `W`.
-/// Use [`MultiGzDecoder`] if your file has multiple streams.
 ///
```
joshtriplett, line 174:
```suggestion
/// This decoder only handles gzipped data with a single stream.
/// Use [`MultiGzDecoder`] for gzip files with multiple streams.
```

src/gz/read.rs
```diff
@@ -180,21 +178,16 @@ impl<R: Read + Write> Write for GzDecoder<R> {
...
-/// specification](https://tools.ietf.org/html/rfc1952), however, allows multiple
-/// gzip members to be joined in a single stream.  `MultiGzDecoder` will
-/// decode all consecutive members while [`GzDecoder`] will only decompress the
-/// first gzip member. The multistream format is commonly used in bioinformatics,
-/// for example when using the BGZF compressed data. It's also useful
-/// to compress large amounts of data in parallel where each thread produces one stream
-/// for a chunk of input data.
+/// A gzip file consists of a series of "members" concatenated one after another.
+/// MultiGzDecoder decodes all members of a file, while [GzDecoder] will only decode
+/// the first member. MultiGzDecoder is preferable in most cases.
```
joshtriplett, line 185:
```suggestion
/// the first member.
```

src/gz/bufread.rs
```diff
@@ -167,11 +167,11 @@ impl<R: BufRead + Write> Write for GzEncoder<R> {
     }
 }
 
-/// A gzip streaming decoder
+/// A decoder for a single member of a gzip file. Prefer [MultiGzDecoder] for
+/// most uses.
```
joshtriplett, line 171:
```suggestion
/// A decoder for a gzip file with a single member.
```

I'd agree that we shouldn't *just* present this as a decoder for gzip in general, and should capture the distinction in the summary, but I'd stop short of telling people to *prefer* MultiGzDecoder.

src/gz/read.rs
```diff
@@ -90,13 +90,11 @@ impl<R: Read + Write> Write for GzEncoder<R> {
     }
 }
 
-/// A gzip streaming decoder
+/// A decoder for a single member of a gzip file. Prefer [MultiGzDecoder] for
+/// most uses.
```
joshtriplett, line 94:
```suggestion
/// A decoder for a gzip file with a single member.
```

I'd agree that we shouldn't *just* present this as a decoder for gzip in general, and should capture the distinction in the summary, but I'd stop short of telling people to *prefer* MultiGzDecoder.

src/gz/bufread.rs
```diff
@@ -397,20 +397,16 @@ impl<R: BufRead + Write> Write for GzDecoder<R> {
     }
 }
 
-/// A gzip streaming decoder that decodes all members of a multistream
+/// A gzip streaming decoder that decodes a complete [gzip file].
```
joshtriplett, line 400:
```suggestion
/// A gzip streaming decoder that decodes a [gzip file] with multiple members.
```

src/gz/read.rs
```diff
@@ -180,21 +178,16 @@ impl<R: Read + Write> Write for GzDecoder<R> {
     }
 }
 
-/// A gzip streaming decoder that decodes all members of a multistream
+/// A gzip streaming decoder that decodes a full [gzip file].
```
joshtriplett, line 181:
```suggestion
/// A gzip streaming decoder that decodes a [gzip file] with multiple members.
```

src/gz/read.rs
```diff
@@ -90,13 +90,11 @@ impl<R: Read + Write> Write for GzEncoder<R> {
...
-/// A gzip streaming decoder
+/// A decoder for a single member of a gzip file. Prefer [MultiGzDecoder] for
+/// most uses.
 ///
 /// This structure exposes a [`Read`] interface that will consume compressed
 /// data from the underlying reader and emit uncompressed data.
-/// Use [`MultiGzDecoder`] if your file has multiple streams.
-///
-/// [`Read`]: https://doc.rust-lang.org/std/io/trait.Read.html
 ///
```
joshtriplett, line 98:
```suggestion
///
/// This decoder only handles gzipped data with a single stream.
/// Use [`MultiGzDecoder`] for gzipped data with multiple streams.
///
```

src/gz/bufread.rs
```diff
@@ -167,11 +167,11 @@ impl<R: BufRead + Write> Write for GzEncoder<R> {
...
 }
 
-/// A gzip streaming decoder
+/// A decoder for a single member of a gzip file. Prefer [MultiGzDecoder] for
+/// most uses.
 ///
 /// This structure consumes a [`BufRead`] interface, reading compressed data
 /// from the underlying reader, and emitting uncompressed data.
-/// Use [`MultiGzDecoder`] if your file has multiple streams.
 ///
```
joshtriplett, line 175:
```suggestion
///
/// This decoder only handles gzipped data with a single stream.
/// Use [`MultiGzDecoder`] for gzipped data with multiple streams.
///
```

joshtriplett:
I added some suggestions implementing my comment https://github.com/rust-lang/flate2-rs/pull/324#issuecomment-1644854362 .

I am *not* attached to exact wording here, feel free to wordsmith. Just going for the general spirit of not pushing people towards MultiGzDecoder.

workingjubilee:
Is there a disadvantage to using MultiGzDecoder when the stream only has a single file?

jsha:
> in many cases I think people will be surprised if they encounter a gzip stream with multiple files in it.

I think people will not be surprised, because their tools do not distinguish a gzip file containing multiple members from a gzip file containing a single member. For instance, here's a gzip file:

[file.txt.gz](https://github.com/rust-lang/flate2-rs/files/12116380/file.txt.gz)

If you run it through `gzip -dc` or `zcat` you will get:

```
$ gzip -dc < file.txt.gz 
jsha is the new administrative assistant
```

However, if you run this Rust program (based on the GzDecoder example), you will get a very incorrect result ("jsha is the new admin") with no error:

```
use flate2::read::GzDecoder;
use std::io::prelude::*;
use std::{fs, io};

fn decode_reader(bytes: &[u8]) -> io::Result<String> {
    let mut gz = GzDecoder::new(&bytes[..]);
    let mut s = String::new();
    gz.read_to_string(&mut s)?;
    Ok(s)
}

fn main() {
    let contents = fs::read("file.txt.gz").unwrap();
    let decoded = decode_reader(&contents).unwrap();
    println!("{}", decoded);
}
```

```
$ cargo run
   Compiling zcat v0.1.0 (/home/jsha/zcat)
    Finished dev [unoptimized + debuginfo] target(s) in 0.57s
     Running `target/debug/zcat`
jsha is the new admin
```

As another concrete example: I came across this issue because I maintain an HTTP client, ureq. We implement automatic gzip decoding. When investigating a bug, I discovered that we were using `GzDecoder` instead of `MultiGzDecoder`. That's wrong; a response with `Content-Encoding: gzip` a gzip file and may very well have multiple members. But it's particularly scary because when that happens, people would get only part of the file, and no error indicating anything was wrong.

So I think we should push people towards MultiGzDecoder. It's doesn't have a "silently do the wrong thing" failure mode. If we were designing from scratch, I would argue that the library shouldn't have a single-stream decoder that wraps `Read` / `BufRead`, because of the likelihood of silently dropping data. Instead, to support use cases that want to know about stream boundaries, there should be a secondary API that yields an `Iterator<Item = GzipMember>`, where `GzipMember` implements read. And then if the user chooses to drop the Iterator after reading a single member that's fine, but it's clear that there _may_ be multiple members to deal with.

Now, I will say this, even though it slightly contradicts my point: In writing up my arguments I also went to check what various HTTP clients do with `Content-Encoding: gzip` and a response that contains multiple members. To my great surprise, the ones I checked seem to fail after the first member. For instance:

```
$ curl https://jacob.hoffman-andrews.com/file.txt --compressed
curl: (23) Failed writing received data to disk/application
jsha is the new admin
```

And Chrome and Firefox both display "jsha is the new admin" without any indication of error. I would still argue that this is a bug, but it is evidently a common one. I know all three programs use zlib; perhaps the API design of zlib encourages this particular bug?

workingjubilee:
I agree with jsha: the evidence that Rust programs which do not adhere to this advisory will drop data on the floor is evidence enough that we should in fact recommend MultiGzDecoder is used. While not memory unsoundness, Rust programmers often are... *irked* when they find out an interface that they were using sends half or more of their data to `/dev/null`.

jongiddy:
There is a difference in behavior that should be noted. `MultiGzDecoder` behaves differently for a gzip member embedded in other data. If there is non-gzip data following the gzip data `GzipDecoder` succeeds, `MultiGzDecoder` fails.

As an example, for the `write` versions `GzipDecoder` returns the number of bytes consumed, leaving any remaining bytes for the caller to handle. `MultiGzDecoder` returns an `Err` since it expects more gzip data.

Correct behavior depends on knowledge of the data. Is it only gzip data and may have multiple members, or is it a stream of data elements that contains gzipped data? HTTP clients have a real problem here as some servers send multi-member gzip responses and some other servers send additional data after the gzip-encoded content. There is probably a server that does both.

That is why it is legitimate to have two types of decoder, and it is the caller who needs to choose which is most appropriate for their data.

src/gz/bufread.rs
jongiddy: Whatever the final text, it should stick to the RFC terminology of "member" rather than using "stream" in some places.

src/gz/write.rs
Byron: I have applied this change here and to the other 2 places where the same text appears.

Byron:
After having read this I feel that instead of nudging towards a particular implementation, it would be preferable to teach about the concept of "members" [on the front-page](https://docs.rs/flate2/1.0.26/flate2/) while providing some of the examples we have seen here to explain what this can mean for an application. From there, users should be able to make the choice for their particular application.

For this PR this could mean one more section in the `lib.rs` documentation, along with the explicit lack of preferences in the wording for the documentation of `Gz` and `MultiGz` variants of types.

If this sounds feasible, give this comment a thumbs up. Otherwise, you could thumbs down (just to have an easy count) along with suggestions on what to do instead. Thanks everyone!

joshtriplett:
@jongiddy wrote:
> `MultiGzDecoder` behaves differently for a gzip member embedded in other data. If there is non-gzip data following the gzip data `GzipDecoder` succeeds, `MultiGzDecoder` fails.

That's one of the reasons, yeah. We should document that as a reason people should consider when choosing what to use.

Also, as @jsha noted, single-member decoding seems to be what many browsers and web clients do, and it'll be surprising if you get different results in different tools. Especially if the subsequent members are silently appended when they may be intended to be different "files".

It'd be worth coordinating with browser vendors to find out whether they have any practical considerations here (e.g. some hard-won knowledge about what happens in the wild). Are there some Firefox developers in the relevant area that we could poke?

@jsha wrote:
> If we were designing from scratch, I would argue that the library shouldn't have a single-stream decoder that wraps Read / BufRead, because of the likelihood of silently dropping data. Instead, to support use cases that want to know about stream boundaries, there should be a secondary API that yields an `Iterator<Item = GzipMember>`,

Yeah, I'd agree with this. We could still add that API, which seems much more reasonable than MultiGzDecoder since it doesn't silently traverse the boundaries between files. And *that* API seems more reasonable to steer people towards, because it doesn't make the same tradeoffs of not being able to continue reading. In particular, if we're careful, you could iterate over the first member and then read the underlying stream for non-gzip data (without any issue at the boundary). (The iterator type could have a "give me the rest of the data as a Read/BufRead instance" method that consumes the iterator.) And conversely, we could have a method on the iterator that consumes the iterator and gives you a single stream but *errors* if there's more data afterwards, so that when people aren't expecting appended data they can check that assumption.

Summarizing:

- Could we check with Firefox and/or Curl developers to find out if there's practical knowledge here that we're lacking? Or, are there open bugs, where knowledge like "this is a bug but here's why we can't fix it" or "this is a bug and we should actually fix it but we haven't yet" might live? If they have specific hard-won considerations about data in the wild, I'd be happy to defer to those. But the fact that *multiple* pieces of software out there *don't* decode multiple streams by default makes me hesitant to diverge.
- In the meantime, it seems entirely reasonable for the documentation to document, at length, the considerations that should go into picking between the options we have. Perhaps we could explain this in detail at the top level documentation of the crate, and then link that from all the various decoders to avoid duplication? This includes explaining the single-stream/multi-stream issue, the tradeoffs about being able to have non-gzip data after one gzip stream, the fact that the single-stream decoder will ignore a subsequent gzip stream, the fact that many pieces of software don't seem to decode multiple streams...
- We might be able to have a better API that better encourages people to explicitly handle this consideration. I *don't* think we should block merging better documentation until we have a better API, but I do think that having a better API would make the issue visible in the type system so that people need to handle it.

workingjubilee:
Hearing there are common use-cases that combine a gz-encoded stream with something else does seem worth thinking about. It feels like people should be introduced to the idea that if they are handling a simple "gz file" that they probably should just handle it with MultiGzDecoder but that if they are handling a more complex encoding they might need to do something more complex and that is where GzDecoder comes into play. Ironically, it is the "simpler" version that is used for the "complex" cases, which violates many intuitions.

jongiddy:
Exactly, and it is also the case that multi-member gzip files are uncommon (outside of Bioinformatics) so `GzDecoder` works for the common use case most of the time, which makes it unexpected when it doesn't.

I wonder whether it is worth introducing new types `GzipDecoder` and `GzipMemberDecoder` that alias `MultiGzDecoder` and `GzDecoder` and then deprecate the `Gz` types? On one hand it fixes the simple vs complex problem, on the other it might just add more confusion.

workingjubilee:
Maybe in the future? It seems like it'd just add more confusion right now. Improving docs and adding the hypothetical iterator seems like the "way to go" in terms of cleaning up the API, probably.

jongiddy:
The iterator idea is a cool abstraction of the problem but not really needed by users of this crate. Almost no-one cares about the individual members, they just want all the uncompressed data from the gzip data whether or not it is multi-member and whether or not it is embedded in other data. The one exception of which I am aware is the [BGZF format](http://samtools.github.io/hts-specs/SAMv1.pdf), where an index is used to seek to the correct gzip member and then uncompress just that one member. Neither benefits from the iterator abstraction.

workingjubilee:
I don't disagree that most users will basically go `.iter().collect::<String>()` or whatever, but it was also feature requested in https://github.com/rust-lang/flate2-rs/issues/192 for "My data is HUGE! Please let me lazily iterate it!" use-cases.

the8472:
> The iterator idea is a cool abstraction of the problem but not really needed by users of this crate. Almost no-one cares about the individual members, they just want all the uncompressed data from the gzip data whether or not it is multi-member and whether or not it is embedded in other data. 

As a low-level crate it still makes sense to expose the basic building blocks for those who need it and then provide an abstraction for the most common use-cases.

workingjubilee:
> along with the explicit lack of preferences in the wording for the documentation of `Gz` and `MultiGz` variants of types.
> 
> If this sounds feasible, give this comment a thumbs up. Otherwise, you could thumbs down (just to have an easy count) along with suggestions on what to do instead. Thanks everyone!

Altogether, for this PR in particular, I'm in favor of this, modulo "we shouldn't be shy about saying that MultiGzDecoder is the simpler option for a 'basic' file". :+1:

> Could we check with Firefox and/or Curl developers to find out if there's practical knowledge here that we're lacking? Or, are there open bugs, where knowledge like "this is a bug but here's why we can't fix it" or "this is a bug and we should actually fix it but we haven't yet" might live? If they have specific hard-won considerations about data in the wild, I'd be happy to defer to those. But the fact that multiple pieces of software out there don't decode multiple streams by default makes me hesitant to diverge.

Let's make this a separate issue, but I suspect this might be a case of "EXTREMELY pedantic adherence to a specific RFC's spec in a way that is unnecessarily underachieving".

workingjubilee:
Also see:
- https://github.com/rust-lang/flate2-rs/issues/339

This issue is an extremely common papercut.

Byron:
I have added a new top-level section to explain the intricacies of the `gzip` file format. Further *I did my best* to integrate all of Josh's suggestions, while also not applying a new paragraph if it appeared to say the same thing as a paragraph above it.

I particularly invite for a thorough review of the new top-level section.

jsha (commented):
This is looking great. Thanks for all the thoughtful conversation and feedback, all. One of the things it's clarified is that the main thing that winds up being risky / surprising is that the single-member variants drop data on the floor; so we should explicitly document that fact on those variants. I'll write up another commit with that change plus some minor tweaks that I just commented, and push that shortly.

jsha:
Okay, pushed a batch of changes. Sorry they wound up bigger than I intended; I was hoping to converge piecemeal, but wound up wanting to try and get all six places using the same language.

> Could we check with Firefox and/or Curl developers to find out if there's practical knowledge here that we're lacking?

This is a good idea. I'm not likely to get it done soon, but I think it would be really informative and interesting.

To make sure I got it all right, copy-pasting from the generated docs for {Read,BufRead,Write}{Gz,MultiGz}Decoder:

[bufread](https://github.com/rust-lang/flate2-rs/pull/index.html)::[GzDecoder](https://github.com/rust-lang/flate2-rs/pull/324#)

A decoder for the first member of a [gzip file](https://www.rfc-editor.org/rfc/rfc1952#page-5).

This structure exposes a [BufRead](https://doc.rust-lang.org/std/io/trait.BufRead.html) interface, reading compressed data from the underlying reader, and emitting uncompressed data.

After reading the first member of a gzip file (which is often, but not always, the only member), this reader will return Ok(0) even if there are more bytes available in the underlying reader. If you want to be sure not to drop bytes on the floor, call into_inner() after Ok(0) to recover the underlying reader.

To handle gzip files that may have multiple members, see [MultiGzDecoder](https://github.com/rust-lang/flate2-rs/pull/struct.MultiGzDecoder.html) or read more [in the introduction](https://github.com/rust-lang/flate2-rs/index.html#about-multi-member-gzip-files).

[read](https://github.com/rust-lang/flate2-rs/pull/index.html)::[GzDecoder](https://github.com/rust-lang/flate2-rs/pull/324#)

A decoder for the first member of a [gzip file](https://www.rfc-editor.org/rfc/rfc1952#page-5).

This structure exposes a [Read](https://doc.rust-lang.org/nightly/std/io/trait.Read.html) interface that will consume compressed data from the underlying reader and emit uncompressed data.

After reading the first member of a gzip file (which is often, but not always, the only member), this reader will return Ok(0) even if there are more bytes available in the underlying reader. If you want to be sure not to drop bytes on the floor, call into_inner() after Ok(0) to recover the underlying reader.

To handle gzip files that may have multiple members, see [MultiGzDecoder](https://github.com/rust-lang/flate2-rs/pull/struct.MultiGzDecoder.html) or read more [in the introduction](https://github.com/rust-lang/flate2-rs/index.html#about-multi-member-gzip-files).

[write](https://github.com/rust-lang/flate2-rs/pull/index.html)::[GzDecoder](https://github.com/rust-lang/flate2-rs/pull/324#)

A decoder for the first member of a [gzip file](https://www.rfc-editor.org/rfc/rfc1952#page-5).

This structure exposes a [Write](https://doc.rust-lang.org/std/io/trait.Write.html) interface, receiving compressed data and writing uncompressed data to the underlying writer.

After decoding the first member of a gzip file, this writer will return XXX to all subsequent writes.

To handle gzip files that may have multiple members, see [MultiGzDecoder](https://github.com/rust-lang/flate2-rs/pull/struct.MultiGzDecoder.html) or read more [in the introduction](https://github.com/rust-lang/flate2-rs/index.html#about-multi-member-gzip-files).

[bufread](https://github.com/rust-lang/flate2-rs/pull/index.html)::[MultiGzDecoder](https://github.com/rust-lang/flate2-rs/pull/324#)

A gzip streaming decoder that decodes a [gzip file](https://www.rfc-editor.org/rfc/rfc1952#page-5) that may have multiple members.

This structure exposes a [BufRead](https://doc.rust-lang.org/std/io/trait.BufRead.html) interface that will consume compressed data from the underlying reader and emit uncompressed data.

A gzip file consists of a series of members concatenated one after another. MultiGzDecoder decodes all members of a file and returns Ok(0) once the underlying reader does.

To handle members seperately, see [GzDecoder](https://github.com/rust-lang/flate2-rs/pull/struct.GzDecoder.html) or read more [in the introduction](https://github.com/rust-lang/flate2-rs/index.html#about-multi-member-gzip-files).

[read](https://github.com/rust-lang/flate2-rs/pull/index.html)::[MultiGzDecoder](https://github.com/rust-lang/flate2-rs/pull/324#)

A gzip streaming decoder that decodes a [gzip file](https://www.rfc-editor.org/rfc/rfc1952#page-5) that may have multiple members.

This structure exposes a [Read](https://doc.rust-lang.org/nightly/std/io/trait.Read.html) interface that will consume compressed data from the underlying reader and emit uncompressed data.

A gzip file consists of a series of members concatenated one after another. MultiGzDecoder decodes all members of a file and returns Ok(0) once the underlying reader does.

To handle members seperately, see [GzDecoder](https://github.com/rust-lang/flate2-rs/pull/struct.GzDecoder.html) or read more [in the introduction](https://github.com/rust-lang/flate2-rs/index.html#about-multi-member-gzip-files).

[write](https://github.com/rust-lang/flate2-rs/pull/index.html)::[MultiGzDecoder](https://github.com/rust-lang/flate2-rs/pull/324#)

A gzip streaming decoder that decodes a [gzip file](https://www.rfc-editor.org/rfc/rfc1952#page-5) with multiple members.

This structure exposes a [Write](https://doc.rust-lang.org/nightly/std/io/trait.Write.html) interface that will consume compressed data and write uncompressed data to the underlying writer.

A gzip file consists of a series of members concatenated one after another. MultiGzDecoder decodes all members of a file and writes them to the underlying writer one after another.

To handle members separately, see [GzDecoder](https://github.com/rust-lang/flate2-rs/pull/struct.GzDecoder.html) or read more [in the introduction](https://github.com/rust-lang/flate2-rs/index.html#about-multi-member-gzip-files).

Byron (approved):
This looks great to me, thanks everyone for their help and contributions!

I'd be fine merging this PR and probably will do so if there was silence here for a week or so, or nobody does it before me 😁.

src/gz/bufread.rs
```diff
@@ -167,12 +167,22 @@ impl<R: BufRead + Write> Write for GzEncoder<R> {
     }
 }
 
-/// A gzip streaming decoder
+/// A decoder for the first member of a [gzip file].
```
jongiddy, line 170: Since GzDecoder is used for formats that provide an index to lookup a single member that may not be the first, I'd prefer this to be
```suggestion
/// A decoder for a single member of a [gzip file].
```

src/gz/bufread.rs
```diff
@@ -167,12 +167,22 @@ impl<R: BufRead + Write> Write for GzEncoder<R> {
...
-/// This structure consumes a [`BufRead`] interface, reading compressed data
+/// This structure exposes a [`BufRead`] interface, reading compressed data
 /// from the underlying reader, and emitting uncompressed data.
-/// Use [`MultiGzDecoder`] if your file has multiple streams.
 ///
+/// After reading the first member of a gzip file (which is often, but not
+/// always, the only member), this reader will return Ok(0) even if there
+/// are more bytes available in the underlying reader. If you want to be sure
+/// not to drop bytes on the floor, call `into_inner()` after Ok(0) to
+/// recover the underlying reader.
```
jongiddy, line 179:
```suggestion
/// After reading a single member of the gzip data this reader will return 
/// Ok(0) even if there are more bytes available in the underlying reader. 
/// If you need the following bytes, call `into_inner()` after Ok(0) to
/// recover the underlying reader.
```

src/gz/bufread.rs
```diff
@@ -397,20 +407,19 @@ impl<R: BufRead + Write> Write for GzDecoder<R> {
...
-/// specification](https://tools.ietf.org/html/rfc1952), however, allows multiple
-/// gzip members to be joined in a single stream. `MultiGzDecoder` will
-/// decode all consecutive members while [`GzDecoder`] will only decompress
-/// the first gzip member. The multistream format is commonly used in
-/// bioinformatics, for example when using the BGZF compressed data. It's also useful
-/// to compress large amounts of data in parallel where each thread produces one stream
-/// for a chunk of input data.
+/// A gzip file consists of a series of *members* concatenated one after another.
+/// MultiGzDecoder decodes all members of a file and returns Ok(0) once the
+/// underlying reader does.
```
jongiddy, line 417:
```suggestion
/// MultiGzDecoder decodes all members from the read data and only returns Ok(0) when the
/// underlying reader does. For a gzip file, this reads to the end of the file.
```

src/gz/read.rs
```diff
@@ -90,13 +90,22 @@ impl<R: Read + Write> Write for GzEncoder<R> {
     }
 }
 
-/// A gzip streaming decoder
+/// A decoder for the first member of a [gzip file].
```
jongiddy, line 93: Same changes as for bufread

src/gz/write.rs
```diff
@@ -166,12 +166,19 @@ impl<W: Write> Drop for GzEncoder<W> {
     }
 }
 
-/// A gzip streaming decoder
+/// A decoder for the first member of a [gzip file].
```
jongiddy, line 169:
```suggestion
/// A decoder for a single member of a [gzip file].
```

src/gz/write.rs
```diff
@@ -166,12 +166,19 @@ impl<W: Write> Drop for GzEncoder<W> {
...
 ///
-/// This structure exposes a [`Write`] interface that will emit uncompressed data
-/// to the underlying writer `W`.
-/// Use [`MultiGzDecoder`] if your file has multiple streams.
+/// This structure exposes a [`Write`] interface, receiving compressed data and
+/// writing uncompressed data to the underlying writer.
+///
+/// After decoding the first member of a gzip file, this writer will return XXX
+/// to all subsequent writes.
```
jongiddy, line 175:
```suggestion
/// After decoding a single member of the gzip data this writer will return the number of bytes read 
/// to the end of the gzip member and subsequent writes will return Ok(0) allowing the caller to 
/// handle any data following the gzip member.
```

src/lib.rs
```diff
@@ -65,12 +65,33 @@
 //! `Write` trait if `T: Write`. That is, the "dual trait" is forwarded directly
 //! to the underlying object if available.
 //!
+//! # About multi-member Gzip files
+//!
+//! While most `gzip` files one encounters will have a single *member* that can be read
+//! with the [`GzDecoder`], there may be some files which have multiple members.
+//!
+//! If these are read with a [`GzDecoder`], only the first member will be consumed and
+//! the rest will silently be left alone, which can be surprising.
```
jongiddy, line 74:
```suggestion
//! A [`GzDecoder`] will only read the first member of gzip data, which may unexpectedly 
//! provide partial results when a multi-member gzip file is encountered. Some data is 
//! designed to be read as single members from a multi-member file. `GzDecoder` also allows 
//! non-gzip data following gzip data to be handled.
```

src/lib.rs
```diff
@@ -65,12 +65,33 @@
...
+//! The [`MultiGzDecoder`] on the other hand will decode all members of a `gzip` file
+//! into one consecutive stream of bytes, which hides the underlying *members* entirely.
+//! If a file contains contains non-gzip data after the gzip data, MultiGzDecoder will
+//! emit an error after decoding the gzip data. This behavior matches the `gzip`,
+//! `gunzip`, and `zcat` command line tools.
+//!
+//! Chrome and Firefox appear to implement behavior like `GzDecoder`, ignoring data
+//! after the first member. `curl` appears to implement behavior somewhat like
+//! `GzDecoder`, only decoding the first member, but emitting an error if there is
+//! data after the first member, whether or not it is gzip data.
```
jongiddy, line 85: This is interesting for determining common practice, but the documentation should describe `flate2` behavior, not that of other tools.

jongiddy:
We can wordsmith endlessly, so these are just suggestions that would clarify things for me. The only one that definitely needs to be changed is the instance that currently says `writer will return XXX`.

src/gz/read.rs
```diff
@@ -90,13 +90,22 @@ impl<R: Read + Write> Write for GzEncoder<R> {
...
 /// This structure exposes a [`Read`] interface that will consume compressed
 /// data from the underlying reader and emit uncompressed data.
-/// Use [`MultiGzDecoder`] if your file has multiple streams.
 ///
-/// [`Read`]: https://doc.rust-lang.org/std/io/trait.Read.html
+/// After reading the first member of a gzip file (which is often, but not
+/// always, the only member), this reader will return Ok(0) even if there
+/// are more bytes available in the underlying reader. If you want to be sure
+/// not to drop bytes on the floor, call `into_inner()` after Ok(0) to
+/// recover the underlying reader.
```
jongiddy, line 102: This doesn't apply to `read::GzDecoder`. The caller loses any bytes between the end of the gzip data and the end of the last read done by the `GzDecoder`. Not losing these bytes is the benefit of using `bufread::GzDecoder`.

```suggestion
/// After reading a single member of the gzip data this reader will return
/// Ok(0) even if there are more bytes available in the underlying reader.
/// `GzDecoder` may have read additional bytes past the end of the gzip data.
/// If you need the following bytes, wrap the `Reader` in a `std::io::BufReader` 
/// and use `bufread::GzDecoder` instead.
```

Byron: Thanks for this clarification, this is subtle.
I must admit that I am still absolutely confused by the many different implementations of what seems to be types with the same name.

src/lib.rs
Byron: I agree and will remove the last paragraph.

Byron:
Thanks a lot, @jongiddy , for pushing this PR over the finishing line.

I have applied all of your suggestions and followed up on all comments as well as I could. Given the massive size of this PR, working with it doesn't get easier and I would like to make the cut here by merging what we have, and invite those interested to double-check what's there to possibly submit follow-up PRs for fixes.

I for one learned a lot here and believe this PR represents a substantial improvement to the docs that will also help users to avoid these member-releated pitfalls (*along with pitfalls related to `GzDecoder` reading a little bit ahead of the underlying stream so one wants to use the `BufRead` implementation to see these nonetheless*).

src/gz/read.rs
jongiddy: Generally the `Read` version is the natural approach for sync code, and the `Write` version is the natural approach for async code.

The `Read` version has the above problem, which the `BufRead` version solves. The `Read` version then delegates to the `BufRead` version. The `Read` and `BufRead` versions could have been the same, but at the cost that a `Read: BufRead` type would get buffered twice, since there is no easy way to convert to a `BufRead` only if `!BufRead`.

What should probably have been better is for the `Read` version to not have the `as_ref`, `as_mut`, and `into_inner` methods to extract the underlying `Read`. Since they do exist, we should add a note that these methods do not work as expected. But that can be a separate PR.

Byron: I have checked again and didn't see this very useful information in the top-level crate documentation, even though I think it definitely has a place there.
If you agree, would you mind sending a PR for this? If there isn't enough time, please let me know as well and I'd take it on me to incorporate this knowledge, stating you as co-author. Thanks a lot.

PR #326, opened 2022-11-23
Add CIFuzz Github action

DavidKorczynski:
Add [CIFuzz](https://google.github.io/oss-fuzz/getting-started/continuous-integration/) workflow action to have fuzzers build and run on each PR.

This is a service offered by OSS-Fuzz where flate2-rs already runs. CIFuzz can help detect regressions and catch fuzzing build issues early, and has a variety of features (see the URL above). In the current PR the fuzzers gets build on a pull request and will run for 300 seconds.

### DIFF
--- a/.github/workflows/cifuzz.yml
+++ b/.github/workflows/cifuzz.yml
@@ -1,0 +1,26 @@
+name: CIFuzz
+on: [pull_request]
+jobs:
+  Fuzzing:
+    runs-on: ubuntu-latest
+    steps:
+    - name: Build Fuzzers
+      id: build
+      uses: google/oss-fuzz/infra/cifuzz/actions/build_fuzzers@master
+      with:
+        oss-fuzz-project-name: 'flate2-rs'
+        dry-run: false
+        language: rust
+    - name: Run Fuzzers
+      uses: google/oss-fuzz/infra/cifuzz/actions/run_fuzzers@master
+      with:
+        oss-fuzz-project-name: 'flate2-rs'
+        fuzz-seconds: 300
+        dry-run: false
+        language: rust
+    - name: Upload Crash
+      uses: actions/upload-artifact@v3
+      if: failure() && steps.build.outcome == 'success'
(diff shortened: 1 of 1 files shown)


thomcc (approved):
This looks fine to me, although I'm not exactly an expert on the the security setup of our repositories.

@Mark-Simulacrum (or, really, anybody on t-infra) is there anything you think we'd need to be concerned about here?

Mark-Simulacrum:
This seems fine, but we may want to update the original cc list for failures - https://github.com/google/oss-fuzz/pull/5599/files#diff-a4e84032c8eeac617821c8f137b2d9c905e295837217ad3cc81960d9d242699cR10

I'm not sure this affects the PR configuration, but presumably affects something. https://github.com/rust-lang/team/blob/master/teams/libs.toml#L43 seems like a reasonable email to use, though we can also define one for crate-maintainers specifically.

JohnTitor (approved):

JohnTitor:
> I'm not sure this affects the PR configuration, but presumably affects something. [rust-lang/team@`master`/teams/libs.toml#L43](https://github.com/rust-lang/team/blob/master/teams/libs.toml?rgh-link-date=2022-11-24T12%3A20%3A11Z#L43) seems like a reasonable email to use, though we can also define one for crate-maintainers specifically.

I'm a bit concerned that sending fuzzing failures emails to t-libs could be noisy, so preparing an alias for crate-maintainers sounds better to me.

Byron (approved):
This PR has been inactive for more than 6 months and I think it's worth taking the leap and merge it.

My reasoning is as follows:

* Google has already been fuzzing this project, with the current email settings, and thus far there have been no complaints about messaging.
* This PR adds just-in-time fuzzing (despite being short) which adds a chance to catch certain kinds of bugs and is thus valuable. The cost of ownership seems acceptable for that.

Finally, if it doesn't work it, the additional fuzzing can easily be removed.

Issue #294, opened 2022-02-18
Examples that don't just read_to_string


SenojLuap:
I have no idea how common the `read_to_string()` use case is, but I've been struggling with errors attempt to decompress to a file, and wouldn't mind seeing some examples of that.

MichaelMcDonnell:
@SenojLuap, I've created an [example that decompresses a file](https://github.com/MichaelMcDonnell/flate2-rs/blob/decompress_file/examples/decompress_file.rs). Is that what you were looking for?

SenojLuap:
Oof. Had to remember what it was I was working on a year ago 😢 
But, I'm pretty sure this is close to the scenario I was interested in, yes.

Thanks!

PR #329, opened 2022-12-16
Add decompress file example

MichaelMcDonnell:
There was already an example of how to compress a file but no example of how to decompress a file. This example more or less copies the compression example but does decompression instead.

It solves issue #294 where someone wanted an example like this.


JohnTitor (approved):
Thank you!

MichaelMcDonnell:
Thanks for merging!

PR #330, opened 2022-12-22
Fixed overflow bug in crc combine

AntonJMLarsson:
The combine function for the CRC hash can cause integer overflow if the amt variable is close to u32::MAX. To fix this I changed the regular add to wrapping add, the same way the update function works.


JohnTitor (approved):
That is what we do on the `update` method, makes sense to me.

PR #336, opened 2023-02-04
Enable all-features, Use doc_auto_cfg on docs.rs

wcampbell0x2a:
- Display functions that are limited by feature, list feature required


JohnTitor (approved):
Makes sense to me, it allows users to choose their preferred backend by looking at docs.rs.

PR #337, opened 2023-02-14
Fix a typo in doc for write::GzDecoder

yestyle:
Hi there,

While reading the documents of flate2-rs, I found a typo in `write::GzDecoder`. It should be emitting uncompressed data if I understand it correctly, so here's the pull request to fix it. Thanks.

Regards,
Philip


JohnTitor (approved):
Thanks!

PR #344, opened 2023-05-01
Move GzHeader into GzState

jongiddy:
Moving the `GzHeader` struct into the `GzState` enum reduces the size of `bufread::GzDecoder<&[u8]>` from 296 to 208 bytes.

This implements part of the change from #323 with the aim to make that PR simpler.


JohnTitor (approved):
Great refactoring, thanks!

PR #351, opened 2023-05-17
Fix a comment on the `Compression` struct

### LINKED ISSUE #350: The documentation for Compression refers to it as "enum"
https://github.com/rust-lang/flate2-rs/blob/fe15e4da08b3433f427cd1d41c6a5ed80a621651/src/lib.rs#LL157C15-L157C15

It is, in fact, not an enum.

JohnTitor:
Fixes #350


Byron (approved):
Thanks a lot for the fix!

PR #363, opened 2023-07-17
Fix header CRC calculation of trailing zeros

jongiddy:
Having said #323 was good, I then found an error: `b"0"` being used where `b"\0"` is needed. It causes a false corrupt error if a gzipped file includes a filename or comment and also has the header CRC.

To test I re-implemented the CRC from the RFC directly. This confirms that the CRC used in the library is correct.

### DIFF
--- a/src/gz/mod.rs
+++ b/src/gz/mod.rs
@@ -182,7 +182,7 @@ impl GzHeaderParser {
                         read_to_nul(r, filename)?;
                         if let Some(crc) = crc {
                             crc.update(filename);
-                            crc.update(b"0");
+                            crc.update(b"\0");
                         }
                     }
                     self.state = GzHeaderState::Comment(crc.take());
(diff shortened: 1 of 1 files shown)


src/gz/mod.rs
```diff
@@ -508,6 +508,85 @@ mod tests {
...
+    fn roundtrip_header() {
+        let mut header = GzBuilder::new()
+            .mtime(1234)
+            .operating_system(57)
+            .filename("filename")
+            .comment("comment")
+            .into_header(Compression::fast());
+
+        // Add a CRC to the header
+        header[3] = header[3] ^ super::FHCRC;
```
Byron, line 569: I wonder if this header-CRC feature is something that should be natively supported one day. Right now I am sure (nearly) nobody would ever do this, even though it seems like a feature people might find interesting to harden their streams.

Maybe that's an easy win for future PRs.

Byron (approved):
Thanks a lot for the fix!

I validated that the tests are failing without the fix in the library code, so all looks good to me.
I am also merging this fix right away on the grounds of its simplicity.

src/gz/mod.rs
jongiddy: While it would be a simple change, there doesn't seem to be demand for it. Neither the `gzip` command nor the Python `gzip` library create a header CRC. Which explains why this bug hasn't caused any problems for real gzipped files.

PR #367, opened 2023-07-30
Document that `read::GzDecoder` consumes bytes after end of gzip

jongiddy:
Add tests showing that the `GzDecoder`s in `bufread` and `write` support reading immediately after end of gzip data.

`read::GzDecoder` can read past the end of gzip data, and a subsequent read of the inner type loses those bytes.

### DIFF
--- a/src/lib.rs
+++ b/src/lib.rs
@@ -117,7 +117,14 @@ mod zlib;
 /// Types which operate over [`Read`] streams, both encoders and decoders for
 /// various formats.
 ///
+/// Note that the `read` decoder types may read past the end of the compressed
+/// data while decoding. If the caller requires subsequent reads to start
+/// immediately following the compressed data  wrap the `Read` type in a
+/// [`BufReader`] and use the `BufReader` with the equivalent decoder from the
+/// `bufread` module and also for the subsequent reads.
+///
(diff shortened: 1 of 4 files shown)


Byron (approved):
Thanks a lot codifying the subtleties that thus far where only kept in issues, while improving the documentation to help users avoid those pitfalls. It's much appreciated!

I have made a few tweaks and believe this PR is good to go.

@jongiddy Could you ping me here in about a week's time so I can hit the merge button? Maybe there is other reviews during that period. (*I will also try to get back here, but maybe I miss the mark*)

jongiddy:
I added a note to the `read` module describing the behaviour of read decoders and how to get the typically expected behaviour.

src/lib.rs
```diff
@@ -117,7 +117,14 @@ mod zlib;
 /// Types which operate over [`Read`] streams, both encoders and decoders for
 /// various formats.
 ///
+/// Note that the `read` decoder types may read past the end of the compressed
+/// data while decoding. If the caller requires subsequent reads to start
+/// immediately following the compressed data  wrap the `Read` type in a
+/// [`BufReader`] and use the `BufReader` with the equivalent decoder from the
+/// `bufread` module and also for the subsequent reads.
```
Byron, line 124: Thank you, I think that's a good, which naturally covers all implementations.

PR #375, opened 2023-09-10
Fix and unify docs of `bufread` and `read` types.

### LINKED ISSUE #374: flate2::bufread::GzDecoder doesn't impl BufRead?
The document says:

https://github.com/rust-lang/flate2-rs/blob/f285e9abac98aa4f0a15d2e79993a261f166056c/src/gz/bufread.rs#L22-L24

but `flate2::bufread::GzDecoder` only implement `Read`.

https://github.com/rust-lang/flate2-rs/blob/f285e9abac98aa4f0a15d2e79993a261f166056c/src/gz/bufread.rs#L281

`BufRead` interface is not implemented and exposed.

georeth:
Goals:
- fix wrong docs of `bufread::GzDecoder`, `bufread::GzEncoder`, `bufread::MultiGzDecoder`.
- make the types of inner reader and exported trait explicit.
- explicit state the pull model: user initialize the read, then data is read from the underlying stream.

Fix #374

I'm not a fluent english speaker and know little about flate2 internals, feel free to edit this PR.

### DIFF
--- a/src/deflate/bufread.rs
+++ b/src/deflate/bufread.rs
@@ -7,9 +7,10 @@ use crate::{Compress, Decompress};
 
 /// A DEFLATE encoder, or compressor.
 ///
-/// This structure consumes a [`BufRead`] interface, reading uncompressed data
-/// from the underlying reader, and emitting compressed data.
+/// This structure implements a [`Read`] interface. When read from, it reads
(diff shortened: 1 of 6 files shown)


georeth:
@Byron

Byron (approved):
Thanks a lot for the fix, it looks good to me - I validated that the implementations match the new advertisement, which indeed is a good match now.

Issue #312, opened 2022-08-23
Decompress Example


EricFecteau:
Receiving zlib raw stream data from python with the "zlib.compressobj(zlib.Z_DEFAULT_COMPRESSION, zlib.DEFLATED, -zlib.MAX_WBITS)" options is not rare, but there are no examples of how to use the "Decompress" object in the example folder and there is no mention of the "Decompress::new_with_window_bits" in the [documentation](https://docs.rs/flate2/latest/flate2/struct.Decompress.html?search=new_with_window_bits). From my understanding, there is no other way to inflate the below example with this library (since the second "message" is dependent on the first). 

Could documentation on the "new_with_window_bits" be added and maybe an example (similar to the one below) be added to the examples?

Python Code for Generating example:

```python
import zlib

compressobj = zlib.compressobj(
    zlib.Z_DEFAULT_COMPRESSION, zlib.DEFLATED, -zlib.MAX_WBITS
)

message = b'{"msgs":[{"msg": "ping"}]}'
compressed = compressobj.compress(message)
compressed += compressobj.flush(zlib.Z_SYNC_FLUSH)
compressed = compressed[:-4]
print([c for c in compressed])

message = b'{"msgs":[{"msg": "lobby_clear"},{"msg": "lobby_complete"}]}'
compressed = compressobj.compress(message)
compressed += compressobj.flush(zlib.Z_SYNC_FLUSH)
compressed = compressed[:-4]
print([c for c in compressed])
```

Rust Code for inflating the above:

```rust
use flate2::{Decompress, FlushDecompress};
use std::str;

fn main() {
    // Python ZLIB compressed with options: zlib.Z_DEFAULT_COMPRESSION, zlib.DEFLATED, -zlib.MAX_WBITS

    // b'{"msgs":[{"msg": "ping"}]}'
    let msg_vec1 = vec![
        170, 86, 202, 45, 78, 47, 86, 178, 138, 174, 6, 49, 148, 172, 20, 148, 10, 50, 243, 210,
        149, 106, 99, 107, 1, 0, 0, 0, 255, 255,
    ];

    // b'{"msgs":[{"msg": "lobby_clear"},{"msg": "lobby_complete"}]}'
    let msg_vec2 = vec![
        170, 198, 144, 201, 201, 79, 74, 170, 140, 79, 206, 73, 77, 44, 82, 170, 213, 65, 23, 206,
        207, 45, 200, 73, 45, 73, 5, 105, 5, 0,
    ];

[...]

Issue #276, opened 2021-07-09
Replace interior Reader without resetting decompress data


jnbooth:
If I'm parsing a continuous encoded stream that is (unfortunately) divided up across several successive inputs, my only option appears to be using the Decompress object directly, which is a huge hassle without access to the zio:: functions.

It seems like it would be easy to create a "replace" method identical to the various "reset" methods, but with the `reset_decoder_data(self);` call removed. That way, I could switch out one Read object for the next while preserving the stream state.

Byron:
I am probably not following on some intricacies of the problem, but wonder if it's possible to create a custom `Write/Read` that is `Clone` and uses shared ownership internally. That way it would be possible to keep a handle to it after passing it to types in this crate, and use interior mutability to swap out any state as you see fit.

I'd find that, in theory, more suitable as first step towards solving this particular problem as it's much less involved. Of course, it also delegates the hard parts of managing the internal decompressor state correctly which certainly wouldn't be trivial, but should be possible.

nc-x:
Tried swapping out the internal reader for GzDecoder using interior mutability but it does not work.

- You have to swap when the first internal reader is completely read 
- However, when the first reader is completely read, you get an error: `corrupt deflate stream` (because the stream reader only had partial data).
- In this case, ignoring the error and swapping the reader does nothing as GzDecoder is internally in an error state and it goes into a infinite loop.
- Also, if there was a way to know the decoded size of a reader, we could have read_exact it. But knowing it in advance is impossible, so...

I don't think this can be done without additional support from the library itself.

Byron:
Couldn't a multi-member (continuous) stream be accomplished by a [MultiGzDecoder](https://docs.rs/flate2/latest/flate2/read/struct.MultiGzDecoder.html).

I think what would really help here is an example program with real input data.

nc-x:
AFAICT, MultiGzDecoder is for gzip having multiple files inside it which is not the usecase here. The usecase here, for example, is to download huge gzip files using [range requests](https://developer.mozilla.org/en-US/docs/Web/HTTP/Range_requests). Reqwest's Response object implements Read, so it can be used with GzDecoder or MultiGzDecoder, but once a given response is decoded you make another range request to get the next piece of data and decode it in a streaming fashion, probably because the data is huge and you want to filter out some stuff or parse it directly etc.

But because every range request yields a different response, an ability to swap the internal reader is required while keeping the internal state sane.

[...]

PR #376, opened 2023-09-13
Add `continue_with` function to ZlibDecoder that allows refreshing of the stream without resetting the bufreader and fix resetting of ZlibDecoder

PierreV23:
Examples:
```rust
use flate2::{Decompress, read::ZlibDecoder};
use std::io::prelude::Read;

fn continue_with_example() {
    // Python ZLIB compressed with options: zlib.Z_DEFAULT_COMPRESSION, zlib.DEFLATED, -zlib.MAX_WBITS

    // b'{"msgs":[{"msg": "ping"}]}'
    let msg_vec1 = vec![
        170, 86, 202, 45, 78, 47, 86, 178, 138, 174, 6, 49, 148, 172, 20, 148, 10, 50, 243, 210,
        149, 106, 99, 107, 1, 0, 0, 0, 255, 255,
    ];

    // b'{"msgs":[{"msg": "lobby_clear"},{"msg": "lobby_complete"}]}'
    let msg_vec2 = vec![
        170, 198, 144, 201, 201, 79, 74, 170, 140, 79, 206, 73, 77, 44, 82, 170, 213, 65, 23, 206,
        207, 45, 200, 73, 45, 73, 5, 105, 5, 0,
    ];

    let mut decompresser = ZlibDecoder::new_with_decompress(
        &msg_vec1[..],
        Decompress::new_with_window_bits(false, 15),
    );

    let mut msg1 = String::new();
    decompresser.read_to_string(&mut msg1).unwrap();
    println!("{}", msg1);

    decompresser.continue_read(&msg_vec2[..]);

    let mut msg2 = String::new();
    decompresser.read_to_string(&mut msg2).unwrap();
    println!("{}", msg2);
}

fn reset_example() {
    // b'{"msgs":[{"msg": "ping"}]}'
    let msg_vec1 = vec![
        170, 86, 202, 45, 78, 47, 86, 178, 138, 174, 6, 49, 148, 172, 20, 148, 10, 50, 243, 210,
        149, 106, 99, 107, 1, 0, 0, 0, 255, 255,
    ];

    let mut decompresser = ZlibDecoder::new_with_decompress(
        &msg_vec1[..],
        Decompress::new_with_window_bits(false, 15),
    );

    let mut msg1 = String::new();
    decompresser.read_to_string(&mut msg1).unwrap();
    println!("{}", msg1);

    decompresser.reset_on_custom(&msg_vec1[..], false);

    let mut msg2 = String::new();
    decompresser.read_to_string(&mut msg2).unwrap();
    println!("{}", msg2);
}

```

This should help #312 and #276

NOTE: This does not affect read&write ZlibEncoder and write ZlibDecoder. These are yet to be written.

### DIFF
--- a/src/zlib/bufread.rs
+++ b/src/zlib/bufread.rs
@@ -215,6 +215,16 @@ impl<R> ZlibDecoder<R> {
         &mut self.obj
     }
 
+    /// Acquires a reference to the underlying `Decompress` instance
+    pub fn get_data(&self) -> &Decompress {
+        &self.data
+    }
+
+    /// Acquires a mutable reference to the underlying `Decompress` instance
+    pub fn get_data_mut(&mut self) -> &mut Decompress {
+        &mut self.data
+    }
+
     /// Consumes this decoder, returning the underlying reader.
     pub fn into_inner(self) -> R {
         self.obj
--- a/src/zlib/bufread.rs
+++ b/src/zlib/bufread.rs
@@ -48,9 +48,9 @@ impl<R: BufRead> ZlibEncoder<R> {
         }
     }
 
-    /// Creates a new encoder with given `compresson` settings which will
+    /// Creates a new encoder with the given `compression` settings which will
     /// read uncompressed data from the given stream `r` and emit the compressed stream.
-    pub fn new_with_compress(r: R, compression: crate::Compress) -> ZlibEncoder<R> {
+    pub fn new_with_compress(r: R, compression: Compress) -> ZlibEncoder<R> {
         ZlibEncoder {
             obj: r,
             data: compression,
@@ -176,13 +176,11 @@ impl<R: BufRead> ZlibDecoder<R> {
     }
 
     /// Creates a new decoder which will decompress data read from the given
-    /// stream.
-    ///
-    /// Also takes in a Decompress instance.
-    pub fn new_with_decompress(r: R, decompress: Decompress) -> ZlibDecoder<R> {
+    /// stream, using the given `decompression` settings.
+    pub fn new_with_decompress(r: R, decompression: Decompress) -> ZlibDecoder<R> {
         ZlibDecoder {
             obj: r,
-            data: decompress,
+            data: decompression,
         }
     }
 }
--- a/src/zlib/read.rs
+++ b/src/zlib/read.rs
@@ -222,6 +222,17 @@ impl<R> ZlibDecoder<R> {
         self.inner.get_mut().reset(r)
     }
 
+
+    /// Resets the state of the decoder entirely
+    /// Opposed to `reset`, this function takes into account that there was
+    /// previously a custom Decompress in place
+    /// 
+    /// `zlib_header` should be whatever was used before in `new_with_decompress`
+    pub fn reset_on_custom(&mut self, r: R, zlib_header: bool) -> R {
+        self.inner.get_data_mut().reset(zlib_header);
+        self.inner.get_mut().reset(r)
+    }
+
     /// Resets the stream for another, without wiping the bufreader.
     /// Use this when your 2nd stream is dependant on previous' streams.
     pub fn continue_read(&mut self, r: R) -> R {
--- a/src/zlib/read.rs
+++ b/src/zlib/read.rs
@@ -222,11 +222,10 @@ impl<R> ZlibDecoder<R> {
         self.inner.get_mut().reset(r)
     }
 
-
     /// Resets the state of the decoder entirely
     /// Opposed to `reset`, this function takes into account that there was
     /// previously a custom Decompress in place
-    /// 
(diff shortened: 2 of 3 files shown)


Byron (commented):
Thanks a lot for tackling this, even though I would have loved a [conversation around alternatives](https://github.com/rust-lang/flate2-rs/issues/276#issuecomment-1717083083) first.

The way this library is structured, any method added as to be duplicated *and* adapted 6 times at least, which adds to the maintenance burden.

Solving a problem without changing the API of these types is thus preferred, and I believe that's possible here. My preference here would be to add a utility type which can be used to do these input-swaps.

PierreV23:
I edited the the pull request. (I didn't mean to pull both `continue_with` and the reset fix at the same time, im a newbie...)

Yes, I agree a API change might be needed, but I was afraid of changing too much and I didn't want to change already existing code.

PierreV23:
The biggest issue at the moment is the fact that `ZlibDecoder::reset` uses `bufread::reset_decoder_data`.
```rust
pub fn reset_decoder_data<R>(zlib: &mut ZlibDecoder<R>) {
    zlib.data = Decompress::new(true);
}
```

It recreates an entirely new `Decompress` instance without taking into account possible custom parameters (zlib_header and window_bits). This causes deflate errors. `reset_with_custom` can be seen as a replacement.

`continue_with` allows for a refreshment of the stream, without touching the `Decompress` instance

Byron (changes requested):
Thanks for the examples, they are certainly a straightforward way of solving this particular issue. Due to the cost in maintenance that this additional API surface would incur, I'd like to investigate alternatives.

Can you instead try to create a `Read` implementation that allows to change an internal reader? Think along the lines of `struct Foo(Rc<RefCell<impl std::io::Read>>)`. Then one can clone `Foo` to pass it as reader to the decompressor, but also call `foo.reset(new_reader)` on it. This has the same effect as this API, but is entirely under control of the user.

If properly implemented, tested and documented, I think it could start out as part of an `example`, similar to the one you already provide.

Thanks for your understanding and for considering continuing this PR despite the change in direction.

PierreV23:
> Thanks for the examples, they are certainly a straightforward way of solving this particular issue. Due to the cost in maintenance that this additional API surface would incur, I'd like to investigate alternatives.
> 
> Can you instead try to create a `Read` implementation that allows to change an internal reader? Think along the lines of `struct Foo(Rc<RefCell<impl std::io::Read>>)`. Then one can clone `Foo` to pass it as reader to the decompressor, but also call `foo.reset(new_reader)` on it. This has the same effect as this API, but is entirely under control of the user.
> 
> If properly implemented, tested and documented, I think it could start out as part of an `example`, similar to the one you already provide.
> 
> Thanks for your understanding and for considering continuing this PR despite the change in direction.

So if I understood you correctly, a new implementation of `Read` gets passed to `bufreader::ZlibDecoder`. Where you can directly manipulate the new `Read` instead manipulating it video ZlibDecoder?

Byron:
Yes. You can implement it as part of an official `example/*.rs` which could be what you presented here, but solving it with a `Read` wrapper that allows to swap out the inner `Read` instance.

PierreV23:
That would mean we would not need a reset function for any of the top layer wrappers right? I was thinking of how to remove some of the boilerplate I came by. What if we just make Config struct(s) that are mandatorily passed to ::new? This would also mean we could abandon the separate gz and zlib directories, generalizing them into a single Decoder and Encoder.

We can use a Config struct and make default instances like ZLIB_DEFAULT and GZ_DEFAULT, or, Config::zlib_default() and Config::gz_default().


I think there is a lot of code that can practically be discarded, with the result of creating a new interface that is prettier, more implicit and easier to use.

It would mean a move to a new major version though.

Byron:
It's great that you are taking the time to get acquainted, this will definitely be helpful!

> That would mean we would not need a reset function for any of the top layer wrappers right?

I don't know what that means, but recommend trying to do as described by me earlier. There is no need to adjust anything in this crate to get your example to work.

> It would mean a move to a new major version though.

Any breaking change could be explored in an own crate, and maybe from there one could conclude that it's worth making such a change here as well. This needs a lot of experimentation, time and discussion though to assure it's the right move, it has great implications.

**I think to get this PR merged it has to pivot towards a utility type and an example [as described earlier](https://github.com/rust-lang/flate2-rs/pull/376#pullrequestreview-1624368849)**, and I urge to stay on topic or close this PR and work on a new crate that shows how to do better than what you find here. Thanks for your understanding.

PierreV23:
I think I will drop working on this now then. I am not experienced enough with Read to implement such things and I'd rather work on something else (i.e. generalized / config system).

I will take a break for now and look back once I have enough spare time again.

Good luck.

PR #400, opened 2024-04-16
add `zlib-rs` support via the `libz-rs-sys` C api for `zlib-rs`

folkertdev:
This PR adds (experimental) support for using flate2 with the zlib-rs backend. The `zlib-rs` crate is a pure-rust implementation of the zlib api that makes use of SIMD to achieve performance close to zlib-ng (currently 5% to 10% slower).

- [ISRG project page](https://www.memorysafety.org/initiative/zlib/)
- [`libz-rs-sys` docs](https://docs.rs/libz-rs-sys/0.1.0/libz_rs_sys/index.html)

For the moment, the libz-rs-sys C api is used for the integration. At some point in the future, zlib-rs will get a more convenient rust api that will require less (hopefully no) unsafe code.

cc @joshtriplett

### DIFF
--- a/.github/workflows/main.yml
+++ b/.github/workflows/main.yml
@@ -43,6 +43,8 @@ jobs:
       if: matrix.build != 'mingw'
     - run: cargo test --features zlib-ng --no-default-features
       if: matrix.build != 'mingw'
+    - run: cargo test --features zlib-rs --no-default-features
+      if: matrix.build != 'mingw'
     - run: cargo test --features cloudflare_zlib --no-default-features
       if: matrix.build != 'mingw'
     - run: |
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -19,6 +19,7 @@ and raw deflate streams.
 [dependencies]
 libz-sys = { version = "1.1.8", optional = true, default-features = false }
 libz-ng-sys = { version = "1.1.8", optional = true }
+libz-rs-sys = { version = "0.1.1", optional = true, default-features = false, features = ["std", "rust-allocator"] }
 cloudflare-zlib-sys = { version = "0.3.0", optional = true }
 miniz_oxide = { version = "0.7.1", optional = true, default-features = false, features = ["with-alloc"] }
 crc32fast = "1.2.0"
@@ -38,6 +39,7 @@ zlib = ["any_zlib", "libz-sys"]
 zlib-default = ["any_zlib", "libz-sys/default"]
 zlib-ng-compat = ["zlib", "libz-sys/zlib-ng"]
 zlib-ng = ["any_zlib", "libz-ng-sys"]
+zlib-rs = ["any_zlib", "libz-rs-sys"]
 cloudflare_zlib = ["any_zlib", "cloudflare-zlib-sys"]
 rust_backend = ["miniz_oxide", "any_impl"]
 miniz-sys = ["rust_backend"] # For backwards compatibility
--- a/src/ffi/c.rs
+++ b/src/ffi/c.rs
@@ -1,10 +1,8 @@
 //! Implementation for C backends.
-use std::alloc::{self, Layout};
 use std::cmp;
-use std::convert::TryFrom;
 use std::fmt;
 use std::marker;
-use std::os::raw::{c_int, c_uint, c_void};
+use std::os::raw::{c_int, c_uint};
 use std::ptr;
 
 use super::*;
@@ -52,14 +50,28 @@ impl Default for StreamWrapper {
                 reserved: 0,
                 opaque: ptr::null_mut(),
                 state: ptr::null_mut(),
-                #[cfg(all(feature = "any_zlib", not(feature = "cloudflare-zlib-sys")))]
-                zalloc,
-                #[cfg(all(feature = "any_zlib", not(feature = "cloudflare-zlib-sys")))]
-                zfree,
-                #[cfg(not(all(feature = "any_zlib", not(feature = "cloudflare-zlib-sys"))))]
-                zalloc: Some(zalloc),
-                #[cfg(not(all(feature = "any_zlib", not(feature = "cloudflare-zlib-sys"))))]
-                zfree: Some(zfree),
+                #[cfg(all(
+                    feature = "any_zlib",
+                    not(any(feature = "cloudflare-zlib-sys", feature = "libz-rs-sys"))
+                ))]
+                zalloc: allocator::zalloc,
+                #[cfg(all(
+                    feature = "any_zlib",
+                    not(any(feature = "cloudflare-zlib-sys", feature = "libz-rs-sys"))
+                ))]
+                zfree: allocator::zfree,
+
+                #[cfg(all(feature = "any_zlib", feature = "cloudflare-zlib-sys"))]
+                zalloc: Some(allocator::zalloc),
+                #[cfg(all(feature = "any_zlib", feature = "cloudflare-zlib-sys"))]
+                zfree: Some(allocator::zfree),
+
+                // for zlib-rs, it is most efficient to have it provide the allocator.
+                // The libz-rs-sys dependency is configured to use the rust system allocator
+                #[cfg(all(feature = "any_zlib", feature = "libz-rs-sys"))]
+                zalloc: None,
+                #[cfg(all(feature = "any_zlib", feature = "libz-rs-sys"))]
+                zfree: None,
             })),
         }
     }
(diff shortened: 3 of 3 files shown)


src/ffi/c.rs
```diff
@@ -52,14 +50,28 @@ impl Default for StreamWrapper {
...
+                zalloc: Some(allocator::zalloc),
+                #[cfg(all(feature = "any_zlib", feature = "cloudflare-zlib-sys"))]
+                zfree: Some(allocator::zfree),
+
+                // for zlib-rs, it is most efficient to have it provide the allocator.
+                // The libz-rs-sys dependency is configured to use the rust system allocator
+                #[cfg(all(feature = "any_zlib", feature = "libz-rs-sys"))]
+                zalloc: None,
+                #[cfg(all(feature = "any_zlib", feature = "libz-rs-sys"))]
+                zfree: None,
```
folkertdev, line 74: the trick is that because zlib-rs can do equality comparisons on the function pointer, it can know when its built-in allocator is used, and this allocator guarantees 64-byte alignment. The allocation of `zfree` and `zalloc` below is not needed (and that implementation is less efficient to deal with allignment).

src/ffi/c.rs
```diff
@@ -75,54 +87,65 @@ impl Drop for StreamWrapper {
     }
 }
 
-const ALIGN: usize = std::mem::align_of::<usize>();
+#[cfg(all(feature = "any_zlib", not(feature = "libz-rs-sys")))]
```
folkertdev, line 90: in the case of zlib-rs, the allocator logic is unused. That is an error for this crate, so it's only conditionally compiled now.

src/ffi/c.rs
```diff
@@ -75,54 +87,65 @@ impl Drop for StreamWrapper {
...
-        let ptr = alloc::alloc(layout) as *mut usize;
-        if ptr.is_null() {
-            return ptr as *mut c_void;
+    fn align_up(size: usize, align: usize) -> usize {
+        (size + align - 1) & !(align - 1)
+    }
+
+    pub type AllocSize = libz::uInt;
+
+    extern "C" fn zalloc(_ptr: *mut c_void, items: AllocSize, item_size: AllocSize) -> *mut c_void {
```
folkertdev, line 106: The only two occurences of this `AllocSize` type are in the signature of `zalloc`, which is a private function. Can this type alias be inlined into the signature?

src/ffi/c.rs
```diff
@@ -410,13 +440,14 @@ mod c_backend {
...
     pub use libz::Z_SYNC_FLUSH as MZ_SYNC_FLUSH;
-    pub type AllocSize = libz::uInt;
 
     pub const MZ_DEFAULT_WINDOW_BITS: c_int = 15;
 
     #[cfg(feature = "zlib-ng")]
     const ZLIB_VERSION: &'static str = "2.1.0.devel\0";
-    #[cfg(not(feature = "zlib-ng"))]
+    #[cfg(feature = "zlib-rs")]
+    const ZLIB_VERSION: &'static str = "0.1.0\0";
```
folkertdev, line 449: zlib-rs has a `const fn zlibVersion() -> *const c_char`. So maybe these constants could all have type `*const c_char`? the only downside I see is that that is harder to print for debugging, but because these are internal constants that seems fine.

.github/workflows/main.yml
```diff
@@ -43,6 +43,8 @@ jobs:
       if: matrix.build != 'mingw'
     - run: cargo test --features zlib-ng --no-default-features
       if: matrix.build != 'mingw'
+    - run: cargo test --features zlib-rs --no-default-features
+      if: matrix.build != 'mingw'
```
folkertdev, line 47: we should eventually be able to compile on mingw, but the current release doesn't, and we don't test for it, so disabling for now.

src/ffi/c.rs
```diff
@@ -75,54 +87,63 @@ impl Drop for StreamWrapper {
...
+    use std::convert::TryFrom;
+    use std::os::raw::c_void;
+
+    const ALIGN: usize = std::mem::align_of::<usize>();
+
+    fn align_up(size: usize, align: usize) -> usize {
+        (size + align - 1) & !(align - 1)
+    }
+
+    pub extern "C" fn zalloc(_ptr: *mut c_void, items: uInt, item_size: uInt) -> *mut c_void {
```
Byron, line 104: What's the motivation for the change from `AllocSize` to `uInt`?

I see this was previously set to `pub type AllocSize = libz::uInt;`, and the `*` imports don't make it easier to see where the type is actually coming from.

src/ffi/c.rs
Byron: It would certainly be nice to automatically obtain the correct value for this, as it would be my guess that these constants quickly go stale otherwise. I also don't know how important these values are, assuming they might be printed under certain circumstances.

Until that's clear, it's probably best to integrate with what's there as is done here.

Byron (commented):
Thanks a lot for contributing what could one day be the new default backend, and replace `minizoxide` in this space.

As CI is failing on Windows, I think that might be a good reason to run tests on Windows (and maybe more) platforms as well.

Further, when looking at the dependencies of `libz-rs-sys`, I was surprised that `zlib-rs` [has what looks like test-dependencies](https://github.com/memorysafety/zlib-rs/blob/main/zlib-rs/Cargo.toml#L21) as part of its normal dependencies. It's clear that these are optional and behind features, but I would have thought that `dev-dependencies` work fine for unit-tests as well. It somehow worried me that this (what seems like a) workaround was necessary, and I hoped there is another way. Maybe APIs could be made public behind a hidden feature toggle, and a separate test-only crate could carry all the dependencies? Then the separation would be clear, and maybe some test-specific code could be extracted into a separate space. This comment is probably not relevant to this PR though, merely a side-note.

I would definitely wait for CI to be fixed, and a second reviewer.

src/ffi/c.rs
folkertdev: it's important that the major version is the same, (that is what is checked), so the value won't go stale that quickly. Also tests will fail hard when that happens, so there is no real risk of the value going state on the main branch here.

src/ffi/c.rs
folkertdev: yeah, `AllocSize` was only used in this signature (turns out) and I didn't think that the alias really had any value at that point.

folkertdev:
thanks for the review

- we're working on windows support (already merged into our `main` branch but needs a little cleanup still). Should be ready soon.
- the dependency situation is kind of nasty. From what I observe, it seems that if crate A has a dev-dependency on crate B, then crate B is not compiled with `cfg(test)` to be true. Our crate B needs to expose/derive things for testing in crate A. So, we haven't yet found a better solution than the feature flags, but it is far from ideal, I'd really like to not be doing that.
- we've been in regular contact with @joshtriplett so he'd be a natural second reviewer here.

Byron:
My apologies, that callout what unnecessary, and lacked the appreciation that is due for it as the backend which *just works*. I got carried away by the pure-Rust-powered zlib-ng like performance as default here, it's something that `gitoxide` benefits from a lot.

But maybe this could be a starting point for eventually bringing both projects together (*without having any idea if this is technically or even legally possible*).

oyvindln:
Ah sorry, maybe I came across as overly grumpy, wasn't meant for you or anyone in particular, it's just happened a few times now

Byron:
Right, thanks for clarifying this! If miniz-oxide is 100% safe Rust, there is value in that alone, independently of performance. Then, even if `zlib-rs` is equivalent functionally and regarding portability, figuring out what the default backend should be wouldn't be straightforward anymore and probably stay as is :).

folkertdev:
We certainly have no ill will versus miniz-oxide: it's simple, small, consistent between platforms, no unsafe besides some C api stuff, and performs slightly better than the original zlib, which is perfectly adequate for many use cases.

`zlib-rs` is more ambitious though: it aims to match zlib-ng performance and properly provide the full zlib api, meaning that we can create a drop-in replacement for the zlib dynamic library.

I'm not sure a generic, shared deflate implementation works, but we're happy to share code/knowlegde. It is our impression however that the majority of performance improvements stem from the explicit use of SIMD instructions, which miniz-oxide does not seem to want to use?

---

We believe to have fixed the windows issues. I'll try that here on CI now by setting the dependency to a git branch again. This should not be merged in that state. I'll report back after we've pushed another release to crates.io and think this is ready.

oyvindln:
>It is our impression however that the majority of performance improvements stem from the explicit use of SIMD instructions, which miniz-oxide does not seem to want to use?

We don't want to use any unsafe code in miniz_oxide (external C api requires it of course but that's a separate thing), and as far as I know at least there isn't any way to use explicit SIMD without unsafe on stable rust (the [portable simd](https://doc.rust-lang.org/std/simd/index.html) API could help solve that but it seems to have been stalled as of now). There may be some crates out there that implement something similar though those again will be using unsafe internally.

To what extent explicit SIMD is needed to reach performance parity and to what extent it can be achieved with autovectorization with the right code layout I don't know.

In any case zlib-rs will now provide a rust alternative to using the C implementation of zlib-ng that's close to the same performance for users that are okay with the safety tradeoffs (and that hopefully won't be patched out by debian maintainers like [zlib-ng is](https://salsa.debian.org/rust-team/debcargo-conf/-/blob/master/src/flate2/debian/patches/disable-zlib-ng.patch))

folkertdev:
The explicit SIMD is essential, e.g. https://github.com/memorysafety/zlib-rs/blob/main/zlib-rs/src/deflate/compare256.rs#L170, LLVM just can't see that idea with the wide comparison and bit counting.

There are a couple of tricks around breaking dependency chains in the bit reader that you may benefit from. E.g. https://github.com/Frommi/miniz_oxide/blob/master/miniz_oxide/src/inflate/core.rs#L423 the `if` here is very prone to misprediction. The optimized implementations just always add more bits to the buffer, e.g. we https://github.com/memorysafety/zlib-rs/blob/main/zlib-rs/src/inflate/bitreader.rs#L103 have no branches in that function at all. A bounds check would be acceptable because it is easy to predict.

What's the story behind removing `zlib-ng` from the `flate2` debian package?

folkertdev:
We have released version 0.1.1 that fixes various issues on windows and macos (x86 and aarch64). These targets are now tested in our CI so there should be no regressions. 

From my side this PR is now done and ready to be merged

Byron (approved):
Thanks a lot for all your hard work and for seeing it through!

The only concern I have is that this CI is testing more platforms than the `zlib-rs` CI (even though Windows checks were added now), and I hope that this isn't prone to unexpected breakage here.

Then, I will leave it for @joshtriplett to finalize the review and possibly merge as well.

folkertdev:
What platforms do we miss? mingw is deliberately omitted for now. I guess we could add a build with the beta target?

https://github.com/memorysafety/zlib-rs/blob/main/.github/workflows/checks.yaml#L25

Byron:
Actually, I confused this repo with the one from `libz` which builds on many, many different platforms. Here it's indeed only `mingw` that was omitted. I think my argument *could* still apply, but much more weakly.

I don't think that nightly should play a role as nightly shouldn't ever break anything (and if it does, it's a bug that will be fixed before it hits stable).

joshtriplett (approved):
Looks good to me! Wrote a comment about the feature usage, but I don't think that's a blocker to merging.

Cargo.toml
```diff
@@ -19,6 +19,7 @@ and raw deflate streams.
 [dependencies]
 libz-sys = { version = "1.1.8", optional = true, default-features = false }
 libz-ng-sys = { version = "1.1.8", optional = true }
+libz-rs-sys = { version = "0.1.1", optional = true, default-features = false, features = ["std", "rust-allocator"] }
```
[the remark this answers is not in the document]
folkertdev: yes exactly. it also clarifies that we're explicitly using the rust allocator; that will always be the default when the crate is used from rust, but not when we build a C dynamic library.

folkertdev:
Thanks! would it be possible to make a release with these changes?

Byron:
Yes, cold you create a PR that bumps the version? I can't self-approve my own PRs hence the need for help :).

Byron:
A new release is now available: https://github.com/rust-lang/flate2-rs/releases/tag/1.0.29

torokati44:
Does the README not mention this because it's still experimental (and therefore shouldn't be used "in production")?

Byron:
That's a good point - it might have been an oversight, and it should probably be mentioned along with a note of being not necessarily production-proven yet.

Issue #401, opened 2024-04-23
Continue reading a stream after ZlibDecoder streams finishes


marxin:
I'm implementing the parsing of the git pack file format as part of the coding challenge:
https://app.codecrafters.io/courses/git/stages/7

It seems the git pack format is a binary file format where each object contains a header followed by a Zlib compressed stream. What's unpleasant one doesn't know the size of the **compressed block**. Is it possible to get back the underlying stream (with `into_inner` or `get_mut`) including the buffer data by `ZlibDecoder` so that I can carry on reading another object header?

jongiddy:
#402 adapts the `gzip` test to demonstrate that this does also work for `deflate` and `zlib` `BufRead` decoders.

marxin:
Thank you very much for the fast response! It's great the current `bufread::ZlibDecoder` works as I needed.
I can confirm it works for me in my particular test-case.

Have 2 comments:
- Would it be possible to document the behavior here: https://docs.rs/flate2/latest/flate2/bufread/struct.ZlibDecoder.html#method.into_inner ? Plus I would also make a caveat at `read::ZlibDecoder::into_inner` that one can't do the same.
- Just out of curiosity: why does e.g. `bufread::ZlibDecoder` does not implement `std::io::BufRead` (would be handy as one does not have to wrap the `bufread::ZlibDecoder` again into a `BufReader::new()`?

Byron:
As the original question was answered with tests, I think it's fair to close this issue despite inviting for continuing the conversation here.

Regarding documentation, please feel free to open a PR with the improvement to the docs that you would have wanted to see. Maybe you can play around with `ZlibDecoder` and implementing `BufRead` on it as well. Maybe even more improvements arise from that :).

jongiddy:
There is an [existing discussion](https://github.com/rust-lang/flate2-rs/discussions/385) on why the `bufread` decoders do not implement `BufRead`.

jongiddy:
The docs for bufread and write `GzDecoder` have text describing this behaviour. This can be copied to the docs for the other decoders.

jongiddy:
bufread: https://github.com/rust-lang/flate2-rs/blob/8a502a791fbcbdb56b20f6d6dcd7096f0c8f1a33/src/gz/bufread.rs#L171-L174
[...]

PR #402, opened 2024-04-23
Add tests to show BufRead can be used after decoding

jongiddy:
Copy the trailing data tests from `gzip::bufread` and `gzip::write` to demonstrate that they also work for `zlib` and `deflate`.

This demonstrates that #401 works with `bufread` decoders.

### DIFF
--- a/src/deflate/bufread.rs
+++ b/src/deflate/bufread.rs
@@ -243,3 +243,50 @@ impl<W: BufRead + Write> Write for DeflateDecoder<W> {
         self.get_mut().flush()
     }
 }
+
+#[cfg(test)]
+mod test {
+    use crate::bufread::DeflateDecoder;
+    use crate::deflate::write;
+    use crate::Compression;
+    use std::io::{Read, Write};
+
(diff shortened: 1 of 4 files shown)


Byron (approved):
That's a great idea, and a very valid expectation on top of that.
Thank you!

Byron (changes requested):
Actually, I just realized that these tests should be on integration level (i.e. in `tests/`), they don't use internal APIs (and shouldn't have to).
Thank you.

jongiddy:
I put the tests here for symmetry with the equivalent gzip tests, which are also currently unit tests. While unit tests can access private items, I've never heard that unit tests **must** access private items. And, due to the additional build costs of integration tests, I tend to prefer unit tests where possible.

Nevertheless, I can move the tests to the integration level. Should I move the equivalent gzip tests too?

Byron:
> I put the tests here for symmetry with the equivalent gzip tests, which are also currently unit tests.

Apologies, I wasn't aware and didn't want to change the way things are done here. It seems integration tests have a specific purpose (here, making use of fixtures) and unit-tests are used for everything else. Let's keep it that way.

Byron (approved):

PR #403, opened 2024-04-26
release version 1.0.29: support for zlib-rs

folkertdev:
as requested in #400

PR #405, opened 2024-04-26
fix CI - `--all-features` wasn't tested and didn't work with arrival of `zlib-rs`

### LINKED ISSUE #404: docs.rs failed to build flate2-1.0.29
https://docs.rs/crate/flate2/latest

Byron:
- **fix CI by assuring builds work with --all-features enabled (#404)**
- **CI verifies that docs can be built (#404)**

Fixes #404
Related to #400 and #403

### DIFF
--- a/.github/workflows/main.yml
+++ b/.github/workflows/main.yml
@@ -55,13 +55,14 @@ jobs:
       if: matrix.build == 'stable' 
 
   rustfmt:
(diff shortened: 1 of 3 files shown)


joshtriplett (approved):
Looks good to me. Also a really good argument for using something like `cfg-if` to simplify these conditionals in the future.

PR #408, opened 2024-04-28
docs: Document expected behavior when Read is done for ZLIB and DEFLATE decoders

marxin:
Thanks @jongiddy for guidenance.


Byron (approved):
Thanks so much for this wonderful addition to the documentation! I particularly like how the text considers the differences between the different types.

PR #419, opened 2024-08-11
update maintenance guide with recent news

Byron:
This might be interesting to our new maintainer,  @jongiddy - welcome  🎉.

### DIFF
--- a/MAINTENANCE.md
+++ b/MAINTENANCE.md
@@ -1,5 +1,12 @@
 This document explains how to perform the project's maintenance tasks.
 
+### Repository Configuration
+
+Some repository settings, including branch protection and required approvals, are managed in the
+[Rust team repo](https://github.com/rust-lang/team/blob/master/repos/rust-lang/flate2-rs.toml).
+See the [repository schema](https://github.com/rust-lang/team/blob/aaa8039a2420a07cab9bbaca3885b3ad95a256a7/docs/toml-schema.md?plain=1#L248)
+for supported directives.
+
 ### Creating a new release
 
 #### Artifacts
@@ -11,11 +18,11 @@ This document explains how to perform the project's maintenance tasks.
 
 To generate all the artifacts above, one proceeds as follows:
 
-1. `git checkout -b release-<next-version>` - move to a branch to prepare making changes to the repository. *Changes cannot be made to `main` as it is protected.*
-2. Edit `Cargo.toml` to the next package version.
-3. `gh pr create` to create a new PR for the current branch and **get it merged**.
+1. `git checkout -b release-<next-version>` - create a branch to make changes to the repository. *Changes cannot be made to `main` as it is protected.*
+2. Edit `Cargo.toml` to update the package version to `<next-version>`.
+3. Create a new PR for the branch (e.g. using `gh pr create`) and **get it approved and merged**.
+4. Pull the `main` branch and ensure that you are at the merged commit.
 4. `cargo publish` to create a new release on `crates.io`.
 5. `git tag <next-version>` to remember the commit.
 6. `git push --tags` to push the new tag.
 7. Go to the newly created release page on GitHub and edit it by pressing the "Generate Release Notes" and the `@` button. Save the release.
-


MAINTENANCE.md
```diff
@@ -1,13 +1,23 @@
...
+
 ### Creating a new release
 
 #### Artifacts
 
 * a tag of the version number
 * a new [crate version](https://crates.io/crates/flate2/versions)
 
-#### Process
+#### Process: Plain `cargo publish`
```
jongiddy, line 20: I would like an introductory paragraph to describe why there are two release processes, and which one to use by default.
```suggestion
#### Process
We are moving to using `cargo smart-release` for releases, but it still needs some polishing. [Some criteria to choose which one to use.]
##### `cargo publish`
...
##### `cargo smart-release`
...
```

MAINTENANCE.md
```diff
@@ -19,3 +29,20 @@ To generate all the artifacts above, one proceeds as follows:
...
+However, it's quite new and needs some polishing.
+
+To just update the changelog, run `cargo changelog --write`.
+
+To create a new release, do as follows:
+
+1. `git checkout -b release-<next-version>` - move to a branch to prepare making changes to the repository. *Changes cannot be made to `main` as it is protected.*
+2. `cargo changelog --write` to update the changelog. Edit it to your liking.
+3. Create a commit like `git commit -am "update changelog prior to release"`
+4. `cargo smart-release -e --update-crates-index` - double-check that the release version is sensible, and if not, just edit `Cargo.toml` by hand before.
```
jongiddy, line 44: Does this create a second commit or update the commit from step 3?

jongiddy, line 44:
```suggestion
4. `cargo smart-release -e --update-crates-index` - double-check that the release version is sensible, and if not, edit `Cargo.toml` to specify a sensible release.
```

MAINTENANCE.md
```diff
@@ -19,3 +29,20 @@ To generate all the artifacts above, one proceeds as follows:
...
+5. Tags were created, and they should be pushed, but it might be that it fails to create a GitHub release.
```
jongiddy, line 45: Is this due to use of `v...` tag? Any way to configure `smart-release` to use tags without the `v`?

MAINTENANCE.md
```diff
@@ -19,3 +29,20 @@ To generate all the artifacts above, one proceeds as follows:
...
+
+To create a new release, do as follows:
+
+1. `git checkout -b release-<next-version>` - move to a branch to prepare making changes to the repository. *Changes cannot be made to `main` as it is protected.*
+2. `cargo changelog --write` to update the changelog. Edit it to your liking.
+3. Create a commit like `git commit -am "update changelog prior to release"`
+4. `cargo smart-release -e --update-crates-index` - double-check that the release version is sensible, and if not, just edit `Cargo.toml` by hand before.
+5. Tags were created, and they should be pushed, but it might be that it fails to create a GitHub release. 
+6. Create an additional release tag with the `v` removed, i.e. `git tag -s <version> v<version>`. Then `git push  --tags` by hand.
+7. Double-check the newly created release on GitHub is populated - if not, generate the changelog manually.
```
jongiddy, line 47: It looks like the release for 1.0.31 was not populated. Can you describe the steps more clearly here, perhaps by doing them for 1.0.31?

jongiddy:
In the `cargo publish` section there is an instruction to `get it merged` which implies getting approval to merge. There should be something similar in the `cargo smart-release` section. In both sections, maybe it should read `get the PR approved and merged`.

MAINTENANCE.md
Byron: Actually, this could be up to discussion if having two processes even makes sense.
One day I just 'felt' very strongly about wanting to use `cargo smart-release` but didn't think of the incompatibility regarding the tags, and the other problems that might arise from a PR-based workflow which `cargo smart-release` isn't designed around (so `gh` invocations might fail).

Even though Changelogs are nice, in the end, it might be easier to remove it and call it a failed experiment - I'd actually vote for that.

Byron:
Thanks for the review, @jongiddy , I am very happy to not be left alone here anymore :).

My first suggestion is to remove the remnants of the `cargo smart-release` experiment and go back to doing it like previously described. If you agree, I invite you to make the changes you'd find useful directly and push them to this branch - that's easier and faster as well.

MAINTENANCE.md
jongiddy: Reverting the changes from the Process header down would be clearer. For changelogs, is it possible to add `cargo changelog` to the `cargo publish` workflow?

Byron: That’s definitely possible.

Byron:
Now that you are maintainer @jongiddy , you will probably be able to push into this branch to complete it.
I guess it's my way of saying "hi, and welcome" :).

jongiddy:
I removed the smart-release section and revised the text of the other sections.

Byron (commented):
LGTM! Please feel free to approve and merge.

jongiddy (approved):

jongiddy:
@Byron The repo has auto-merge set, so the PR is automatically merged by you when any maintainer approves. That setting may not make as much sense with multiple maintainers.

Byron:
I think that might have been set for this PR only, I wouldn't expect that to happen next time unless it's explicitly enabled on a per PR basis.

PR #420, opened 2024-08-12
update miniz_oxide dependency to 0.8.x

oyvindln:
Version bump is mainly to change from the [adler](https://crates.io/crates/adler) crate to the [adler2](https://crates.io/crates/adler2) fork of it as the original is no longer maintained 

see https://github.com/rustsec/advisory-db/issues/1992

Also bumps minimum rust version a little but it's still way below what flate2 states as it's minimum supported version.


Byron (approved):
Thanks a lot, much appreciated!

PR #421, opened 2024-08-16
Check minimal version of Rust that compiles

jongiddy:
Specify a minimum version of the Rust compiler that successfully builds the crate. This allows us to be aware if the minimal version increases and to avoid issues such as https://github.com/rust-lang/flate2-rs/issues/370

The minimum version is a best-effort measured value and is different to the MSRV which is a support guarantee. The minimum version can be incremented by a PR in order to pass tests, as long as the MSRV holds. When the minimum version increases, the next release should be a minor version, to allow any affected users to pin to a previous minor version.

### DIFF
--- a/.github/workflows/main.yml
+++ b/.github/workflows/main.yml
@@ -7,8 +7,7 @@ jobs:
     runs-on: ${{ matrix.os }}
     strategy:
       matrix:
-        # I don't really understand the build matrix here...
-        build: [stable, beta, nightly, macos, windows, mingw]
+        build: [stable, beta, nightly, minimal, macos, windows, mingw]
         include:
           - build: stable
             os: ubuntu-latest
@@ -48,11 +47,11 @@ jobs:
     - run: cargo test --features cloudflare_zlib --no-default-features
       if: matrix.build != 'mingw'
     - run: |
-        if ! cargo check --no-default-features 2>&1 | grep "You need to choose"; then 
+        if ! cargo check --no-default-features 2>&1 | grep "You need to choose"; then
           echo "expected message stating a zlib backend must be chosen"
           exit 1
         fi
-      if: matrix.build == 'stable' 
+      if: matrix.build == 'stable'
 
   rustfmt:
     name: Rustfmt & Docs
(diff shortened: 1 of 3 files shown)


jongiddy:
The minimum version for the updated `miniz_oxide` is higher than I expected. Since the `rust-version` in `Cargo.toml` can prevent Cargo from selecting the crate, it needs to be the minimum version supported by a common feature-set. We do not want to prevent someone using Rust 1.56.1 from downloading the crate if they are planning to use `zlib-*` feature.

My current thinking is to remove the `rust-version` directive from `Cargo.toml` but keep the test, and use it to document behavior. (e.g. if using Rust < 1.60 a different backend must be selected).

jongiddy:
Ah it's `ppv-lite86` which is a transitive dependency in the dev-dependencies. v0.2.17 in my lock file has no dependencies, but the latest v0.2.20 used in GitHub Actions adds `zerocopy` which requires 1.60.
```
[dev-dependencies]
├── quickcheck v1.0.3
│   └── rand v0.8.5
│       ├── libc v0.2.153
│       ├── rand_chacha v0.3.1
│       │   ├── ppv-lite86 v0.2.17
```

jongiddy:
Changed the CI workflow to only test that code builds, without running the tests.

README.md
```diff
@@ -23,9 +23,18 @@ flate2 = "1.0"
...
 
-This crate supports the current stable and the last stable for the latest version.
-For example, if the current stable is 1.64, this crate supports 1.64 and 1.63.
-Older stables may work, but we don't guarantee these will continue to work.
+This crate supports the current stable and the last stable compiler for the latest version.
+For example, if the current stable is 1.80, this crate supports 1.80 and 1.79.
+
+Older stables may work, but failures may not be treated as a `flate2` bug.
+
+The `Cargo.toml` file specifies a minimum version for which builds passed at some point.
```
Byron, line 31:
```suggestion
The `Cargo.toml` file specifies a minimum version for which builds passed at some point using the `rust-version` field.
```

Byron: Since 'minimum version' is really the `rust-version` field, I wonder if that name should rather be used to reduce the level of abstraction.

Byron (changes requested):
Thanks a lot for making clearer the defacto way the MSRV was handled.

I'd also hope that adding the `rust-version` can't break anyone. In theory, those who use  even more ancient compiler versions will benefit from those not respecting that field yet. So it should be safe especially with a value as low as that.

There is just one improvement that I think could be made to the MSRV paragraph, which would have helped me to understand it more quickly.

Byron (approved):
Thanks a lot!

.github/workflows/main.yml
```diff
@@ -7,8 +7,7 @@ jobs:
     runs-on: ${{ matrix.os }}
     strategy:
       matrix:
-        # I don't really understand the build matrix here...
-        build: [stable, beta, nightly, macos, windows, mingw]
+        build: [stable, beta, nightly, minimal, macos, windows, mingw]
```
Byron, line 10: CI now fails [on `main`](https://github.com/rust-lang/flate2-rs/actions/runs/10439240311), and I have a feeling that this entry might be the cause of it?

It's surprising that it didn't fail CI before.

In any case, your help for a fix will be welcome.

jongiddy:
Oh wow, CI should complain more noisily if that happens!

PR #422, opened 2024-08-18
Remove non-existent build in CI

jongiddy:
`minimal` build was mistakenly left in the previous commit.


Byron (approved):
Thanks a lot!

And I double-checked: the PR that introduced it succeeded CI with this 'build' value, and even though I saw it, I thought then it must be right and part of these sometimes more complex matrix setups.
Meantime I convinced myself that I somehow merged a failing CI run, but apparently it's not that easy.

PR #424, opened 2024-08-23
increase minimal versions of `libz-ng-sys` and `libz-sys` to their latest releases.

Byron:
That way, those compiling with `-Zminimal-versions` have higher chances of it to work.

See https://github.com/Byron/gitoxide/pull/1541 for reference.

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -19,8 +19,8 @@ and raw deflate streams.
 exclude = [".*"]
 
 [dependencies]
-libz-sys = { version = "1.1.8", optional = true, default-features = false }
-libz-ng-sys = { version = "1.1.8", optional = true }
+libz-sys = { version = "1.1.19", optional = true, default-features = false }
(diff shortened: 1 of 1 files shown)


Byron:
Plan is to put this commit into #425, and prep a new release from there.

Byron:
Closing in favor of #425  .

jongiddy:
I don't understand the problem here. Are earlier versions of these dependencies not working with `flate2`? I would not expect a library to need to raise versions unless it starts to use a new feature of the higher versions or, in some cases, to remove the possibility of using serious bugs discovered in earlier versions.

Byron:
The problem is that the versions specified in `Cargo.toml` are minimal versions that should be able to compile with a compiler at MSRV level.
However, when using `-Zminimal-versions`, it doesn't actually work, which is believed to be fix-worthy. I agree with that, even though it's not very common to operate on the lower bound of compatible dependency versions.

PR #426, opened 2024-08-25
Add minimum build test back in to CI

jongiddy:
This test ensures that the crate builds with the default features for the compiler version listed in `rust-version`.


Byron (approved):
Thank you, I understand now :).

PR #428, opened 2024-09-26
Upgrade zlib-rs to 0.3.0 to get multiple bugfixes

Shnatsel:
Fixes at least these correctness issues:

https://github.com/memorysafety/zlib-rs/pull/165
https://github.com/memorysafety/zlib-rs/issues/169
https://github.com/memorysafety/zlib-rs/issues/172

And possibly others.

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -1,7 +1,7 @@
 [package]
 name = "flate2"
 authors = ["Alex Crichton <alex@alexcrichton.com>", "Josh Triplett <josh@joshtriplett.org>"]
-version = "1.0.33"
+version = "1.0.34"
 edition = "2018"
 license = "MIT OR Apache-2.0"
 readme = "README.md"
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -21,7 +21,7 @@ exclude = [".*"]
 [dependencies]
 libz-sys = { version = "1.1.20", optional = true, default-features = false }
 libz-ng-sys = { version = "1.1.16", optional = true }
-libz-rs-sys = { version = "0.2.1", optional = true, default-features = false, features = ["std", "rust-allocator"] }
+libz-rs-sys = { version = "0.3.0", optional = true, default-features = false, features = ["std", "rust-allocator"] } # this matches the default features, but we don't want to depend on the default features staying the same
 cloudflare-zlib-sys = { version = "0.3.0", optional = true }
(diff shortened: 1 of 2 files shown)


Cargo.toml
```diff
@@ -21,7 +21,7 @@ exclude = [".*"]
 [dependencies]
 libz-sys = { version = "1.1.20", optional = true, default-features = false }
 libz-ng-sys = { version = "1.1.16", optional = true }
-libz-rs-sys = { version = "0.2.1", optional = true, default-features = false, features = ["std", "rust-allocator"] }
```
Byron, line 24: Is there a good reason for removing the explicit handling of features? It looks like [they didn't change](https://docs.rs/crate/libz-rs-sys/latest/features).

The rationale for having them is to not rely on default feature remaining stable but instead take note consciously on which features are enabled.
I'd prefer to keep that.

Byron (changes requested):
Thanks a lot!

Could you also add GitHub links (essentially the body of the PR) into the commit?

Shnatsel:
I believe the merge commit would have PR description as the commit message anyway. But sure, done.

Cargo.toml
```diff
@@ -21,7 +21,7 @@ exclude = [".*"]
...
+libz-rs-sys = { version = "0.3.0", optional = true, default-features = false, features = ["std", "rust-allocator"] } # this matches the default features, but we don't want to depend on the default features staying the same
```
Byron, line 24:
```suggestion
# this matches the default features, but we don't want to depend on the default features staying the same
libz-rs-sys = { version = "0.3.0", optional = true, default-features = false, features = ["std", "rust-allocator"] }
```

Byron (commented):
Thanks a lot, this should do it.

If merge-commits would come up in code-lens or similar tools, that should be enough as well. I am not sure they do though.

Would you consider applying the suggestion? Once again I am too afraid to do it myself as it might prevent merging 😅.

Shnatsel:
I assume @Byron will bump it as part of the usual release process. If not, I'm happy to increment that number, sure.

Byron:
Actually, changing the version number will make a release more straightforward. It's a bit like being a genie, limitless power, but impossible to use for yourself :D.

Byron (approved):

Shnatsel:
Version bumped

Byron (approved):
Thanks a lot!

PR #429, opened 2024-09-26
Revert "Merge pull request #427 from ByteBaker/main"

Shnatsel:
This reverts #427


Byron (changes requested):
Thanks a lot for helping.

However, I don't support reverting the commit, as it's purpose is to explain the 'typically'. The problem with the state it's currently in is that it doesn't warn about `10` not being zlib compatible.

`flate2` users are allowed to use their backends to the fullest extent, and they can be trusted to make their own choices. Previously this was implicit, now it's explicit.

Issue #434, opened 2024-10-06
flate2::Decompress::decompress consumes more input than actually used


link2xt:
[`flate2::Decompress::decompress`](https://docs.rs/flate2/1.0.34/flate2/struct.Decompress.html#method.decompress) is documented as "consuming only as much input as needed and writing as much output as possible". However in my tests it consumes the whole input even if output buffer size is not sufficient. Because of this I cannot use `flate2` to decode the IMAP stream incrementally using [`async-compression`](https://crates.io/crates/async-compression) crate: https://github.com/async-email/async-imap/pull/112

Here is a failing test where output buffer is only 8 bytes long, but the whole input is consumed:
```rust
#[test]
fn deflate_decoder_partial() {
    // Decompresses to
    // "* QUOTAROOT INBOX \"User quota\"\r\n* QUOTA \"User quota\" (STORAGE 76 307200)\r\nA0001 OK Getquotaroot completed (0.001 + 0.000 secs).\r\n"
    let input = vec![
        210, 82, 8, 12, 245, 15, 113, 12, 242, 247, 15, 81, 240, 244, 115, 242, 143, 80, 80, 10,
        45, 78, 45, 82, 40, 44, 205, 47, 73, 84, 226, 229, 210, 130, 200, 163, 136, 42, 104, 4,
        135, 248, 7, 57, 186, 187, 42, 152, 155, 41, 24, 27, 152, 27, 25, 24, 104, 242, 114, 57,
        26, 24, 24, 24, 42, 248, 123, 43, 184, 167, 150, 128, 213, 21, 229, 231, 151, 40, 36, 231,
        231, 22, 228, 164, 150, 164, 166, 40, 104, 24, 232, 129, 20, 104, 43, 128, 104, 3, 133,
        226, 212, 228, 98, 77, 61, 94, 46, 0, 0, 0, 0, 255, 255,
    ];

    // Create very small output buffer.
    let mut output = vec![0; 8];

    let zlib_header = false;
    let mut decompress = flate2::Decompress::new(zlib_header);

    let flush_decompress = flate2::FlushDecompress::None;
    let status = decompress.decompress(&input, &mut output, flush_decompress).unwrap();
    assert_eq!(status, flate2::Status::Ok);

    // Should not consume everything, there is not enough space in the buffer for the output.
    assert_ne!(decompress.total_in(), input.len() as u64);
}
```

[...]

PR #436, opened 2024-10-07
Add test for partial decompression

link2xt:
This adds the test from #434

### DIFF
--- a/tests/decompress_chunk.rs
+++ b/tests/decompress_chunk.rs
@@ -21,7 +21,11 @@ fn deflate_decoder_partial() {
     loop {
         let prev_out = decompress.total_out();
         let status = decompress
-            .decompress(&input[decompress.total_in() as usize..], &mut output_buf, flush_decompress)
+            .decompress(
+                &input[decompress.total_in() as usize..],
+                &mut output_buf,
+                flush_decompress,
+            )
             .unwrap();
         let output_len = decompress.total_out() - prev_out;
         output.extend_from_slice(&output_buf[..output_len as usize]);


Byron:
The question here is if `StreamEnd` is happening in the libz-backend for instance, which is what I would expect.
Further, I'd add an actual assertion on the output, after all, it's known and by the end everything should have been decompressed.

That way, at least with the non-miniz-oxide implementation there is a working version of this. From there it should be possible to figure out what's happening with the `miniz_oxide` backend.

Thinking about it, `gitoxide` does something similar, but also knows the size of the input buffer. However, I don't think it relies on it.

oyvindln:
StreamEnd happens in either backend if the deflate stream is well formed and the decompressor encounters the end of a block with the last block flag set.

At least in case of zlib, BufError is [returned](https://github.com/madler/zlib/blob/d476828316d05d54c6fd6a068b121b30c147b5cd/inflate.c#L587) if deflate is called with Z_FINISH as flush mode and it encounters the end of the input data but it the last block did not have an end flag. Thus the stream is technically not well formed but I presume it's treated here as not an error since it seems that this is common to encounter in many protocols and treating it as an error in the stream class may cause issues.

miniz_oxide is supposed to behave the same way so if it isn't it's a bug.

As for the choice whether to consume up to the size of the internal buffer, or just as much as would fit in the output I don't really know what's the most "correct behaviour. I have to investigate a bit more whether zlib always does the latter as it seems to at least in this situation or if it depends on some parameter/config. (Also not sure whether C miniz did here, whether it different from zlib or whether it was an accidental change during porting - I suspect the latter.) Can try to alter this behaviour in miniz_oxide if people wish.

oyvindln:
I don't really know whether it would make performance worse or not, would have to test, miniz_oxide needs more performance work in any case. The behaviour could maybe also added as a parameter to the deflate function (though that may require a version bump for API change in miniz_oxide, or maybe added as an alternate function)

Byron:
Thanks everyone!

It seems the only action that can be taken here is to update the documentation to specifically mention how `miniz_oxide` behaves right now. Maybe there could be additions to make clearer how `StreamEnd` and `BufError` are to be understood.

> The behaviour could maybe also added as a parameter to the deflate function (though that may require a version bump for API change in miniz_oxide, or maybe added as an alternate function)

I like the current setup here where calls to backends seem to be the same, without special adaptations depending on the backend. But I might be wrong about that, and if so, `flate2` would certainly make use of the flag to get similar behaviour.

As for this PR, it could certainly be merged with the documentation improvements, and with some clear-text description of what the test is validating. For instance, I would expect that the special handling can be removed once `miniz_oxide` is acting similarly as the other backends.

Does that make sense?

Byron:
Both work, but I think it requires more comments to explain what is tested, and what the code should rather look like one day so eventually it can be adjusted to reach its final form.

Byron:
> I did some testing and it seems the old C miniz backend behaved the same way as current miniz_oxide so while I would have to test the C functions to be 100% sure it does seem like this is a difference in behaviour between how the `z_deflate` function acts due to how the between zlib and it's C miniz counterpart due to the differences in how they are implemented miniz_oxide just continued the miniz implementation.

I see - this issue has always been present even in the `miniz` C implementation which `miniz_oxide` is derived off.

> Need to do some more digging whether it's practical to alter miniz_oxide to limit the output to not write more than the size of the output buffer or whether that would require a substantial redesign of the internals.

I wonder if there is any way to fix this here, or to fix `miniz_oxide` to be able to detect such a case so that callers can reliably work with it. To my mind, it's OK to work a little differently as long as it's still possible to work with status codes, which is probably what callers do most of the time.

Byron:
It seems this PR is stuck and I wonder if we can or should do something about it.

If I recall correctly the issue is that the test has to add a special case for `miniz_oxide` which can't easily be fixed there, but then it's the question if we should have a non-portable test at all.

Thinking about it, maybe there can be two versions of the test to make the distinction clear - one for `miniz_oxide` and another for the rest?

PR #439, opened 2024-11-14
upgrade zlib-rs to version `0.4.0`

folkertdev:
fixes a stack overflow on malicious input https://github.com/trifectatechfoundation/zlib-rs/security/advisories/GHSA-j3px-q95c-9683

### DIFF
--- a/src/ffi/c.rs
+++ b/src/ffi/c.rs
@@ -447,7 +447,7 @@ mod c_backend {
     #[cfg(feature = "zlib-ng")]
     const ZLIB_VERSION: &'static str = "2.1.0.devel\0";
     #[cfg(all(not(feature = "zlib-ng"), feature = "zlib-rs"))]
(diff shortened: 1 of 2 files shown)


src/ffi/c.rs
```diff
@@ -447,7 +447,7 @@ mod c_backend {
     #[cfg(feature = "zlib-ng")]
     const ZLIB_VERSION: &'static str = "2.1.0.devel\0";
     #[cfg(all(not(feature = "zlib-ng"), feature = "zlib-rs"))]
-    const ZLIB_VERSION: &'static str = "1.3.0-zlib-rs-0.3.0\0";
```
folkertdev, line 450: this bump is not strictly needed here (only the `1.3.0` part is relevant, and I believe only the `1` prefix is checked by most zlib libraries (but applications might check more).

PR #441, opened 2025-01-06
Fix cfgs

### LINKED ISSUE #440: zlib-rs and cloudflare_zlib features conflict
If you try to enable both the zlib-rs and cloudflare_zlib features:

```toml
[dependencies]
flate2 = { version = "1.0.35", features = ["zlib-rs", "cloudflare_zlib"] }
```
Then there are some compile errors:
```
error[E0252]: the name `libz` is defined multiple times
   --> /home/jonathan/.cargo/registry/src/index.crates.io-6f17d22bba15001f/flate2-1.0.35/src/ffi/c.rs:412:9
    |
409 |     use libz_rs_sys as libz;
    |         ------------------- previous import of the module `libz` here
...
412 |     use cloudflare_zlib_sys as libz;
    |         ^^^^^^^^^^^^^^^^^^^^^^^^^^^ `libz` reimported here
    |
    = note: `libz` must be defined only once in the type namespace of this module
help: you can use `as` to change the binding name of the import
    |
412 |     use cloudflare_zlib_sys as other_libz;
    |                             ~~~~~~~~~~~~~
...
```
I believe this is can be fixed by establishing precedence between the two flags here: 
https://github.com/rust-lang/flate2-rs/blob/14aec222af2b85485bd3df4578d8d510eca46af7/src/ffi/c.rs#L408-L412

kornelski:
Fixes #440

I've grouped combinations of features for each library as one-liners, so that they can be copy'n'pasted to where they're needed, without having to have a SAT solver in your head.

Unfortunately macros can't wrap struct fields, so `cfg-if` can't help here.

Verified with cargo-hack that all 239 combinations of feature flags compile okay:

```bash
cargo hack --feature-powerset --exclude-no-default-features --skip any_impl,any_zlib check
```

### DIFF
--- a/src/ffi/c.rs
+++ b/src/ffi/c.rs
@@ -50,27 +50,38 @@ impl Default for StreamWrapper {
                 reserved: 0,
                 opaque: ptr::null_mut(),
                 state: ptr::null_mut(),
-                #[cfg(all(
-                    feature = "any_zlib",
-                    not(any(feature = "cloudflare-zlib-sys", feature = "libz-rs-sys"))
+
+                #[cfg(any(
+                    // zlib-ng
+                    feature = "zlib-ng",
+                    // libz-sys
+                    all(not(feature = "cloudflare_zlib"), not(feature = "zlib-ng"), not(feature = "zlib-rs"))
                 ))]
                 zalloc: allocator::zalloc,
-                #[cfg(all(
-                    feature = "any_zlib",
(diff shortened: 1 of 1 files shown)


Byron (approved):
Thanks so much for your help and for [sharing your insights](https://github.com/rust-lang/flate2-rs/issues/440#issuecomment-2573623172) earlier.

This should mean that all currently supported backends remain as they all offer certain benefits.

And when looking at the complexity of the `cfg` attributes, we'd probably also avoid picking up more backends 😅.

PR #444, opened 2025-01-13
update CI to use new wasi target

oyvindln:
might as well make a separate PR to fix this while I'm at it


Byron (approved):
Thanks a lot! I hope I can merge while the roundtrip check is failing so we can take it on separately.

PR #445, opened 2025-01-25
Implement `Clone` for `CompressError` and `DecompressError`

mkrasnitski:
When making my own error types, one of my variants wraps a `flate2::DecompressError`, and that prevents me from deriving `Clone` on my error type without creating a mirror implementation and transforming between the `flate2` type and my copy of it. These two error types are cheap to clone and `Status` already derives `Clone`, too.

I didn't add a `Copy` derive because the `ErrorMessage` struct might be changed to contain a string in the future, which would break the `Copy` implementation.

### DIFF
--- a/src/ffi/c.rs
+++ b/src/ffi/c.rs
@@ -8,7 +8,7 @@ use std::ptr;
 use super::*;
 use crate::mem;
 
-#[derive(Default)]
+#[derive(Clone, Default)]
 pub struct ErrorMessage(Option<&'static str>);
 
 impl ErrorMessage {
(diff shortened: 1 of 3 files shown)


mkrasnitski:
Seems like the `cloudflare_zlib` feature is broken unrelated to my changes.

Byron (approved):
Thanks a lot, I'd also think that doing this should only be helpful downstream, and not cause any complications.

CI break spontaneously not too long ago, and it's still unclear what's even causing this. Time to reproduce this locally I presume, but no need to block this PR on it.

PR #449, opened 2025-02-08
Fix CI

Byron:
CI is failing and in such a way that doesn't reproduce locally,
even when using Ubuntu.

Maybe it's related an out-of-memory situation?

### Ideas

* [ ] ~~bisect using CI - when did it start failing?~~
* [ ] ~~bisect tests by inclusion, exclusion~~
    - it's not related to 'big' roundtrips
* [x] let's just get it to work by not testing failing platforms (cloudflare only).

### DIFF
--- a/.github/workflows/main.yml
+++ b/.github/workflows/main.yml
@@ -45,7 +45,8 @@ jobs:
     - run: cargo test --features zlib-rs --no-default-features
       if: matrix.build != 'mingw'
     - run: cargo test --features cloudflare_zlib --no-default-features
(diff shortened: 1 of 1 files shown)


Byron (commented):
This seems to work, nice!

joshtriplett:
Seems reasonable to me to fix CI first before starting to debug it. 

I personally would treat the cloudflare-zlib support as being in maintenance mode at the moment.

joshtriplett (approved):

PR #450, opened 2025-02-08
feat: replace custom u16 le parser with existent rust method

CosminPerRam:
Replaces the custom implementation with the one already provided in the language, they produce the same assembly: [before](https://godbolt.org/z/Gexrh8WoW) and [after](https://godbolt.org/z/5f5bWG5PK).
I also took the liberty to remove the redundant borrows in the functions usages.

### DIFF
--- a/src/gz/mod.rs
+++ b/src/gz/mod.rs
@@ -163,7 +163,7 @@ impl GzHeaderParser {
                         if let Some(crc) = crc {
                             crc.update(buffer);
                         }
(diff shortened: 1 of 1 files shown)


Byron (approved):
Thanks a lot!
Thanks also for validating it turns out to be exactly the same!

Note that I am merging despite CI being red, #449 addresses this but I can't merge it myself, waiting for review.

PR #451, opened 2025-02-09
Do not use cloudflare-zlib-sys 0.3.4

jongiddy:
`cloudflare-zlib-sys` released a new version at the same time that tests started failing. Rather than stop testing cloudflare_zlib, can we pin to the previous version that succeeds in CI? With this restriction encoded in the `Cargo.toml` we avoid the possibility that this problem might be present outside GitHub CI runners and will cause real code to fail.

### DIFF
--- a/.github/workflows/main.yml
+++ b/.github/workflows/main.yml
@@ -45,8 +45,7 @@ jobs:
     - run: cargo test --features zlib-rs --no-default-features
       if: matrix.build != 'mingw'
     - run: cargo test --features cloudflare_zlib --no-default-features
-      # cloudflare doesn't seem to work on CI anymore for some reason, it's not reproducible locally.
-      if: matrix.build != 'mingw' && matrix.os != 'ubuntu-latest' && matrix.os != 'windows-2022'
+      if: matrix.build != 'mingw'
     - run: |
         if ! cargo check --no-default-features 2>&1 | grep "You need to choose"; then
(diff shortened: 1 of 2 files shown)


joshtriplett:
Pinning like this tends to be a really big hammer that can cause other problems.

If we have to do so, we can, but first, can we report this upstream and see if they can yank the broken version? (And then publish a new one once the issue is addressed.)

Byron:
As a compromise, could 'patching the `cloudflare-zlib-sys` version on CI' be an avenue? That way tests could run while it's made clear what the problem is.
Alternatively, cloudflare could be allowed to fail so we see when it's fixed more easily?

joshtriplett:
It looks like cloudflare's zlib repo has issue turned off.

@kornelski, since you work on that repo, do you know the right place to report this? (Sorry for the ping, but the right place to report was not obvious.)

kornelski:
'm looking into the problem. I've also opened issues on cloudflare-zlib repo.

joshtriplett:
@kornelski Thank you! Would you mind yanking the most recent version, until the issue is resolved?

kornelski:
Yanked

kornelski:
The problem was caused by CPU feature detection that made MSVC use AVX in non-AVX builds. I've released a fix.

joshtriplett (approved):

PR #452, opened 2025-02-10
Increase minimum compiler version to 1.67

jongiddy:


Byron (commented):
Thanks a lot! This seems to fix one out of two CI failures.

The one that's left says:

```
Run cargo build
    Updating crates.io index
error: failed to select a version for the requirement `cc = "^1.1"`
candidate versions found which didn't match: 1.0.99, 1.0.98, 1.0.97, ...
location searched: crates.io index
required by package `cloudflare-zlib-sys v0.3.5`
    ... which satisfies dependency `cloudflare-zlib-sys = "^0.3.5"` of package `flate2 v1.0.35 (/home/runner/work/flate2-rs/flate2-rs)`
Error: Process completed with exit code 101.
```

Could it be that dependency resolution takes the minimum Rustc version in account and can't select the version iof `cc` that it would need?

jongiddy:
Yes, I think this forces us to raise the MSRV to 1.67. That's well within the specified support. Probably worth a minor bump on the next release though.

joshtriplett:
@jongiddy Please go ahead and raise the MSRV in `Cargo.toml` to match. (I don't think it needs a minor bump.)

joshtriplett (approved):

PR #453, opened 2025-02-12
Change internal bounds from `R: Read` to `R: BufRead`

JonathanBrouwer:
Reasons to do this:
* This matches the stronger bounds of all callers
* Calling these functions with a `R` that is not a buffered reader would cause performance issues. These functions call `.bytes()` on the reader which will perform very small reads.
* There will soon be a clippy lint that catches this mistake. https://github.com/rust-lang/rust-clippy/pull/14089
* These are not public functions so this is not a breaking change.

### DIFF
--- a/src/gz/mod.rs
+++ b/src/gz/mod.rs
@@ -120,7 +120,7 @@ impl GzHeaderParser {
         }
     }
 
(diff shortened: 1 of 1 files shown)


Byron (approved):
Thanks a lot! That's a great catch and it will prevent the performance issues you mentioned.

It's good these functions aren't public as well, making this PR an easy merge.

PR #457, opened 2025-02-12
feat: remove explicit default impls

CosminPerRam:
`Crc`: The explicit defined impl was added in [this](https://github.com/rust-lang/flate2-rs/commit/0fa0a7b40b8e022f47b59453adda938e5156ac0f#diff-c6bca753500dc9228a31ff821b89aa10e5f37b8cb8f2fcb4dfe16693e969a1dcR26-R31) commit, not sure if there was a reason for it to not be in the derive, Hasher (crc32fast 1.2.0 at that time) had a [Default impl](https://docs.rs/crc32fast/1.2.0/crc32fast/struct.Hasher.html#implementations).

`GzHeaderState`: [here](https://github.com/rust-lang/flate2-rs/commit/68ba8f6e62c7c0a12e3a62cffa2f1cdf7d1f5f30) to build on versions older than 1.62 (?), not applicable [anymore](https://github.com/rust-lang/flate2-rs/pull/452).

`GzBuilder`: [here](https://github.com/rust-lang/flate2-rs/commit/0fa0a7b40b8e022f47b59453adda938e5156ac0f#diff-c6bca753500dc9228a31ff821b89aa10e5f37b8cb8f2fcb4dfe16693e969a1dcR26-R31) (same commit as `Crc`).

Note: as now, `default` and `new` (on `GzBuilder` and `Crc`) are the same, so I replaced explicit construction with a `default` call, please mention if this is not appropiate.

### DIFF
--- a/src/crc.rs
+++ b/src/crc.rs
@@ -8,7 +8,7 @@ use crc32fast::Hasher;
 /// The CRC calculated by a [`CrcReader`].
 ///
 /// [`CrcReader`]: struct.CrcReader.html
-#[derive(Debug)]
+#[derive(Debug, Default)]
 pub struct Crc {
     amt: u32,
     hasher: Hasher,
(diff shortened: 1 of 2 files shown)


Byron (approved):
Thanks a lot, a great catch!

I also don't know why the derive wasn't use, but think it is the way to go as well.

PR #461, opened 2025-02-13
feat: add Error associated type in zio::Ops to handle multiple errors

CosminPerRam:
[Ops::run](https://github.com/rust-lang/flate2-rs/blob/5c6eaec5915ce9a310ab0c28754acd2963fb2740/src/zio.rs#L18) and [Ops::run_vec](https://github.com/rust-lang/flate2-rs/blob/5c6eaec5915ce9a310ab0c28754acd2963fb2740/src/zio.rs#L24) return `Result<Status, DecompressError>`.
There are 2 implementations of it, in [Decompress](https://github.com/rust-lang/flate2-rs/blob/5c6eaec5915ce9a310ab0c28754acd2963fb2740/src/zio.rs#L58) which everything is all good, and in [Compress](https://github.com/rust-lang/flate2-rs/blob/5c6eaec5915ce9a310ab0c28754acd2963fb2740/src/zio.rs#L32), which because of the trait, results in the usage of some `.unwrap()`'s:
https://github.com/rust-lang/flate2-rs/blob/5c6eaec5915ce9a310ab0c28754acd2963fb2740/src/zio.rs#L40-L47
and
https://github.com/rust-lang/flate2-rs/blob/5c6eaec5915ce9a310ab0c28754acd2963fb2740/src/zio.rs#L48-L55
So, adding an associated type in `Ops` covers both.
Because of this, these unwraps are also handled now:
https://github.com/rust-lang/flate2-rs/blob/5c6eaec5915ce9a310ab0c28754acd2963fb2740/src/zio.rs#L255-L277
(without the last one, that's here for context)
The trait is used only internally, so no breaking changes and only less possible panics (:

### DIFF
--- a/src/deflate/write.rs
+++ b/src/deflate/write.rs
@@ -1,7 +1,7 @@
 use std::io;
 use std::io::prelude::*;
 
-use crate::zio;
+use crate::{zio, CompressError, DecompressError};
 use crate::{Compress, Decompress};
 
 /// A DEFLATE encoder, or compressor.
@@ -28,7 +28,7 @@ use crate::{Compress, Decompress};
 /// ```
 #[derive(Debug)]
 pub struct DeflateEncoder<W: Write> {
-    inner: zio::Writer<W, Compress>,
+    inner: zio::Writer<W, CompressError, Compress>,
 }
 
 impl<W: Write> DeflateEncoder<W> {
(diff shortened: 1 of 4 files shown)


Byron (approved):
That's a great catch, along with a great solution!
Thanks so much!

PR #462, opened 2025-02-13
feat: impl From<Flush> to MZFlush

CosminPerRam:
We would never hit an error in building a `MZFlush`, because all values of [FlushDecompress](https://docs.rs/flate2/1.0.35/flate2/enum.FlushDecompress.html) and [FlushCompress](https://docs.rs/flate2/1.0.35/flate2/enum.FlushCompress.html) (as `i32`) map to [MZFlush](https://docs.rs/miniz_oxide/0.8.4/miniz_oxide/enum.MZFlush.html).
Implementing `From` to it is simpler and better as we get rid of the redundant `.unwrap()`s, casts to `i32` and can see directly the translation.

For more context for the mapping done here, see [MZFlush::new](https://docs.rs/miniz_oxide/0.8.4/src/miniz_oxide/lib.rs.html#69-77).

### DIFF
--- a/src/mem.rs
+++ b/src/mem.rs
@@ -52,9 +52,9 @@ pub enum FlushCompress {
     /// All pending output is flushed to the output buffer, but the output is
     /// not aligned to a byte boundary.
     ///
-    /// All of the input data so far will be available to the decompressor (as
-    /// with `Flush::Sync`. This completes the current deflate block and follows
-    /// it with an empty fixed codes block that is 10 bites long, and it assures
+    /// All input data so far will be available to the decompressor (as with
+    /// `Flush::Sync`). This completes the current deflate block and follows it
+    /// with an empty fixed codes block that is 10 bites long, and it assures
     /// that enough bytes are output in order for the decompressor to finish the
     /// block before the empty fixed code block.
     Partial = ffi::MZ_PARTIAL_FLUSH as isize,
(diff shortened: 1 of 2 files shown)


CosminPerRam:
I also took the liberty to reorder `FlushCompress` variants to be more in line with their integer values (before it was `0, 2, 1, 3, 4` now its `0, 1, 2, 3, 4`).

CosminPerRam:
Also found a spelling mistake and a missing `)`.

src/mem.rs
```diff
@@ -49,6 +49,16 @@ pub enum FlushCompress {
     /// accumulate before producing output in order to maximize compression.
     None = ffi::MZ_NO_FLUSH as isize,
 
+    /// All pending output is flushed to the output buffer, but the output is
+    /// not aligned to a byte boundary.
+    ///
+    /// All input data so far will be available to the decompressor (as with
+    /// `Flush::Sync`). This completes the current deflate block and follows it
+    /// with an empty fixed codes block that is 10 bites long, and it assures
```
Byron, line 57:
```suggestion
    /// with an empty fixed codes block that is 10 bytes long, and it assures
```

Byron: I am aware this typos wasn't introduced here, but we might as well fix it now.

Byron (commented):
Thanks a lot for your continued contributions, they are much appreciated!

I also like the small, focussed PRs.

Let's merge this right after another typo 'bit the dust' :).

Byron (approved):
Thanks a lot, appreciated!

You seem to be studying the entire codebase and I am curious to see what else you will find :).

PR #463, opened 2025-02-13
feat: make GzBuilder::filename and GzBuilder::comment results

CosminPerRam:
To handle possible panics with user-povided input, return a `Result` on [GzBuilder::comment](https://docs.rs/flate2/1.0.35/flate2/struct.GzBuilder.html#method.comment) and [GzBuilder::filename](https://docs.rs/flate2/1.0.35/flate2/struct.GzBuilder.html#method.filename).

We build a [CString](https://doc.rust-lang.org/std/ffi/struct.CString.html#method.new), and if it fails, it returns a [NulError](https://doc.rust-lang.org/std/ffi/struct.NulError.html) which converts to [io::Error as an InvalidInput](https://doc.rust-lang.org/src/std/io/error.rs.html#100-105).

Doing so will cause breaking api changes, this is completely debatable if its worth it or not, so this PR is more like a proposal.

For more context:
The unwraps were added here: https://github.com/rust-lang/flate2-rs/commit/6430b38674655039cbe63e5b64320ff9afec45b8, back in 2015, which makes them 10-years old now.

### DIFF
--- a/examples/gzbuilder.rs
+++ b/examples/gzbuilder.rs
@@ -13,8 +13,8 @@ fn main() {
 fn sample_builder() -> Result<(), io::Error> {
     let f = File::create("examples/hello_world.txt.gz")?;
     let mut gz = GzBuilder::new()
-        .filename("hello_world.txt")
-        .comment("test file, please delete")
+        .filename("hello_world.txt")?
+        .comment("test file, please delete")?
         .write(f, Compression::default());
     gz.write_all(b"hello world")?;
     gz.finish()?;
--- a/src/gz/mod.rs
+++ b/src/gz/mod.rs
@@ -346,20 +346,12 @@ impl GzBuilder {
     }
 
     /// Configure the `filename` field in the gzip header.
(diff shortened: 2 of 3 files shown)


jongiddy:
Changing the public API is unlikely to happen. The panic condition is well-documented and can be tested beforehand if the data is provided by the user.

There is a possibility that new `Result`-returning methods could be added, such as `with_filename` but, if this is done, it would be worth discussing which type is most appropriate. Passing a `CString` directly would work and avoid any errors. But arguably a `PathBuf` is a more appropriate type for a filename. The string is encoded in ISO 8859-1 so enforcing this might also be desirable.

Byron:
Thanks a lot for proposing this.

But, I agree with @jongiddy that there'd be better ways to accomplish this.
As this PR won't be merged, I suggest, also for the sake of discussion, to create a new PR that adds a new method so making this improvement won't be a breaking change, if there is interest.

Thanks for your understanding.

PR #464, opened 2025-02-17
feat: remove redundant if guard on option value match

CosminPerRam:


Byron (approved):
Thanks, a good catch!

PR #465, opened 2025-02-18
feat: remove explicit redundant lifetime

CosminPerRam:
My IDE was complaining about it, its fine to do as the borrow is local only to the function call.


Byron (approved):
Thanks a lot!

PR #480, opened 2025-04-01
More informative README on backends

Shnatsel:
I hope this better describes the trade-offs between the different backends.

### DIFF
--- a/README.md
+++ b/README.md
@@ -67,17 +67,20 @@ fn main() {
 
 ## Backends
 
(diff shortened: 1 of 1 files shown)


Byron (approved):
Thanks a lot! To me this looks like a pretty minimal edit to make significant improvements in clarity.

The edit is also factually correct, as far as I can tell.

Event though I will merge in the interest of simplicity, I invite @oyvindln and @folkertdev to submit changes or corrections as they see fit.

PR #481, opened 2025-04-01
Rewrite outdated backend notes in lib.rs

Shnatsel:
This documentation historically was out of sync with the README. For example, the version currently live on docs.rs doesn't mention the non-conflicting `zlib-ng` feature at all, even though it's been around for years.

Even now the documentation contains outdated statements such as "The compression ratios and performance of each of these feature should be roughly comparable"; these days zlib-ng and zlib-rs are significantly faster than the alternatives.

To avoid similar issues going forward, limit the description to the two most relevant backends, and defer to the README that historically has been kept much more up to date for all the details.

### DIFF
--- a/src/lib.rs
+++ b/src/lib.rs
@@ -24,7 +24,7 @@
 //!   This backend is the fastest, at the cost of some `unsafe` Rust code.
 //!
 //! Several backends implemented in C are also available.
-//! These are useful in the rare cases where you were already using a C implementation
+//! These are useful in case you are already using a specific C implementation
 //! and need the result of compression to be bit-identical.
 //! See the crate's README for details on the available C backends.
 //!
--- a/src/lib.rs
+++ b/src/lib.rs
@@ -17,24 +17,18 @@
 //! backends, controlled through this crate's features:
 //!
 //! * `default`, or `rust_backend` - this implementation uses the `miniz_oxide`
-//!   crate which is a port of `miniz.c` (below) to Rust. This feature does not
-//!   require a C compiler and only requires Rust code.
+//!   crate which is a port of `miniz.c` to Rust. This feature does not
+//!   require a C compiler, and only uses safe Rust code.
 //!
-//! * `zlib-rs` - this implementation utilizes the `zlib-rs` crate, a pure rust rewrite of zlib.
-//!   This backend is faster than both `rust_backend` and `zlib`. However, we did not set it as the
-//!   default choice to prevent compatibility issues.
+//! * `zlib-rs` - this implementation utilizes the `zlib-rs` crate, a Rust rewrite of zlib.
+//!   This backend is the fastest, at the cost of some `unsafe` Rust code.
 //!
-//! * `zlib` - this feature will enable linking against the `libz` library, typically found on most
-//!   Linux systems by default. If the library isn't found to already be on the system it will be
-//!   compiled from source (this is a C library).
+//! Several backends implemented in C are also available.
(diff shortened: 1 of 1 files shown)


src/lib.rs
```diff
@@ -17,24 +17,18 @@
...
-//! and performance of each of these feature should be roughly comparable, but you'll likely want
-//! to run your own tests if you're curious about the performance.
+//!   crate which is a port of `miniz.c` to Rust. This feature does not
+//!   require a C compiler, and only uses safe Rust code.
+//!
+//! * `zlib-rs` - this implementation utilizes the `zlib-rs` crate, a Rust rewrite of zlib.
+//!   This backend is the fastest, at the cost of some `unsafe` Rust code.
+//!
+//! Several backends implemented in C are also available.
+//! These are useful in the rare cases where you were already using a C implementation
```
Byron, line 27: I wouldn't assume these cases are *rare* and turn this into a listing of facts about why the C implementation may be useful. Remove `rare` would already do it for me.

Further, I'd love it if there was a way to mention typical problem with using any backend. What comes to mind is duplicate symbols, and issue even Rust backends have today, maybe along with a hint at workarounds. The trigger for this paragraph was me thinking `zlib-ng-compat` was probably good for that, but I wasn't sure.

Byron (changes requested):
Thanks a lot for making `lib.rs` more approachable!

There is only a word that I think should be taken care of, and a question on how some additional context could help dealing with common issues, something the original text might have done but more implicitly. 

Thanks again.

src/lib.rs
Shnatsel: Based on the experiments I've run it seemed like zlib-rs did not result in duplicate symbol issues. So you should be able to use the zlib-rs backend without worrying about what other C zlib flavors may be present, and the whole "use C to avoid duplicate symbols" angle should no longer be a thing.

I am not 100% certain about that though, it's been a while. I'll double-check.

Bit-identical output and symbol conflicts are the only reasons to use C backends, right?

Shnatsel: Hmm, I cannot actually reproduce the linker errors due to conflicting symbols with my old setup.

@folkertdev do you know if zlib-rs would cause symbol conflicts if you try to link it into a binary that also links zlib?

folkertdev: it would cause conflicts right now, but we've added https://github.com/trifectatechfoundation/zlib-rs/pull/322 and will release a new version soon where there should not be any conflicts so long as `export-symbols` is disabled (which is the default)

Byron:
Thanks a lot!

I think it's good enough to merge now, and we can consider improving it further once the new release of `zlib-rs` lands. It would be my goal to make certain compatibility features in `zlib-rs` very approachable even (or particularly) from `flate2`

Byron (approved):

folkertdev:
>  It would be my goal to make certain compatibility features in `zlib-rs` very approachable even (or particularly) from `flate2`

Do you have anything specific in mind here? E.g. with the symbols it's straightforward I think: flate2 will just not enable the feature flag to export symbols, and if users do want that functionality, they can explicitly depend on libz-rs-sys with the feature enabled (and cargo using the union of features should make that work out). Is there anything else/that I'm missing?

Byron:
That sounds like a plan - mitigating issues from the start is certainly better than writing about them in `lib.rs` :). Thanks, I only have vague ideas about this one, but it looks like we will sort this out with the `zlib-rs` update PR and using `zlib-rs` will be a very compatible choice then.

PR #482, opened 2025-04-01
upgrade zlib-rs to version `0.5.0`

folkertdev:
With this release, by default zlib-rs no longer exports C symbols, and hence can be used in projects that do, via some other route, include those symbols (e.g. via `libz-sys` in openssl).

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -1,7 +1,7 @@
 [package]
 name = "flate2"
 authors = ["Alex Crichton <alex@alexcrichton.com>", "Josh Triplett <josh@joshtriplett.org>"]
-version = "1.1.0"
+version = "1.1.1"
 edition = "2018"
 license = "MIT OR Apache-2.0"
 readme = "README.md"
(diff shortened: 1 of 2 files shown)


Byron (approved):
Thanks a lot! This feels like a big step for improving usability.

Also, it feels magical to me as I always thought that `flate2` would use the exported C symbols to get access to the required functions.
But obviously I have no clue how `flate2` actually works 😅.

Byron:
Here is the new release: https://github.com/rust-lang/flate2-rs/releases/tag/1.1.1

folkertdev:
Nice, thanks!

> Also, it feels magical to me as I always thought that flate2 would use the exported C symbols to get access to the required functions.

That's how e.g. `libz-sys` works. `flate2` just uses normal rust name resolution to call other rust functions.

There are cases where other code depends on the symbols being exported (e.g. this is what's happening with openssl), but that is extremely rare, and should really only happen when you also include a C dependency that needs the symbols. We can still do that, it's just not the default any more.

PR #483, opened 2025-04-01
link Update README.md

romashka-btc:
change old link to a new one.

### DIFF
--- a/README.md
+++ b/README.md
@@ -1,7 +1,7 @@
 # flate2
 
 [![Crates.io](https://img.shields.io/crates/v/flate2.svg?maxAge=2592000)](https://crates.io/crates/flate2)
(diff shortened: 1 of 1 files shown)


Byron:
Thanks!

Could you provide a motivation? It looks like GitHub caches it anyway, and it feels safer to let `docs.rs` decide which 'badge-engine' to use.

romashka-btc:
> Thanks!
> 
> Could you provide a motivation? It looks like GitHub caches it anyway, and it feels safer to let `docs.rs` decide which 'badge-engine' to use.

it;s you new link. it;s redirect link , you can check by your own...

Byron:
I don't think changing the `docs.rs` to the redirect is solving a problem, and that appears to be your motivation.

This I think it's best to not go ahead with this.
Thanks for your understanding.

PR #485, opened 2025-04-27
chore: update deps

CosminPerRam:
No major version changes, mostly patches including fixes and small performance improvements.

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -19,16 +19,16 @@ and raw deflate streams.
 exclude = [".*"]
 
 [dependencies]
-libz-sys = { version = "1.1.20", optional = true, default-features = false }
-libz-ng-sys = { version = "1.1.16", optional = true }
+libz-sys = { version = "1.1.22", optional = true, default-features = false }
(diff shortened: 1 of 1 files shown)


Byron (approved):
Thank you, a good catch!

We might want to consider to use dependabot for this - there is probably no good reason to not keep up with the latest versions.

CosminPerRam:
For sure, will make a PR for this soon.

jongiddy:
For a library crate, isn't the effect of these version updates simply to reduce the options available to dependent crates? 

i.e. previously `flate2` was compatible with any version of `libz-sys` >= 1.1.20, but now it requires at least 1.1.22. If, for any reason, a crate needs to keep `libz-sys` at 1.1.20, it now has a conflict.

I tend to update crate versions only when there is a security or bug fix reason to do so.

Byron:
Thanks @jongiddy for chiming in and making that point.

To hammer home the "upgrades only when there is a need" idea, I think we'd also want to have a `cargo audit` job, probably as cron job, to get a heads up early. 
There could also be a section in the README related to manifest update policies just to make it a little more official and accessible.

PR #487, opened 2025-04-28
feat: solve remaining clippy warnings and add it to CI

CosminPerRam:
Other than quite a few usages of `#[allow(clippy::unnecessary_cast)]` cause clippy can't fully see the full picture, things look fine to me.
Many of my contributions contained clippy warning fixes and this PR solves all of the remaining ones and adds check for them to CI (as errors).

### DIFF
--- a/.github/workflows/main.yml
+++ b/.github/workflows/main.yml
@@ -53,7 +53,7 @@ jobs:
         fi
       if: matrix.build == 'stable'
 
-  fmt_docs_clippy:
+  rustfmt_docs_clippy:
     name: Rustfmt, Docs and Clippy
     runs-on: ubuntu-latest
     steps:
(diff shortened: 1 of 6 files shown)


Byron (approved):
Thanks a lot, much appreciated.

The `#[allow(clippy::unnecessary_cast)]` is a bit puzzling to me, but it's nothing to hold the PR up over.

CosminPerRam:
If we try to cast a declaration to the same type as itself (A) that's an unnecessary cast, this lint tries to prevent stuff like this, but Clippy can't always see the full picture (in our case, different definitions behind feature flags, in a place we use type A and somewhere its type B, in this case Clippy sees only the first case, so it thinks we are making an unnecessary cast but we actually dont.

PR #488, opened 2025-04-30
ci: add audit check

CosminPerRam:
Mentioned in #485, as a result of the discussions there this replaces #486 (dependabot updates).


Byron (approved):
Thanks so much, this will be very useful!

Now we can update dependencies only when there is a vulnerability, which would naturally force the downstream to update as well even if they don't run an audit.

Byron:
@CosminPerRam Maybe you could also submit a PR which reverts the previous update of dependencies, as we had no real need for that.

PR #490, opened 2025-06-06
upgrade zlib-rs to version `0.5.1`

folkertdev:
https://github.com/trifectatechfoundation/zlib-rs/releases/tag/v0.5.1

Fixes some minor bugs and improves performance.

### DIFF
--- a/src/ffi/c.rs
+++ b/src/ffi/c.rs
@@ -468,7 +468,7 @@ mod c_backend {
     #[cfg(feature = "zlib-ng")]
     const ZLIB_VERSION: &str = "2.1.0.devel\0";
     #[cfg(all(not(feature = "zlib-ng"), feature = "zlib-rs"))]
-    const ZLIB_VERSION: &str = "1.3.0-zlib-rs-0.5.0\0";
+    const ZLIB_VERSION: &str = "1.3.0-zlib-rs-0.5.1\0";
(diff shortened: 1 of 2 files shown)


Byron (approved):
Thanks a lot!

src/ffi/c.rs
```diff
@@ -468,7 +468,7 @@ mod c_backend {
     #[cfg(feature = "zlib-ng")]
     const ZLIB_VERSION: &str = "2.1.0.devel\0";
     #[cfg(all(not(feature = "zlib-ng"), feature = "zlib-rs"))]
-    const ZLIB_VERSION: &str = "1.3.0-zlib-rs-0.5.0\0";
+    const ZLIB_VERSION: &str = "1.3.0-zlib-rs-0.5.1\0";
```
Byron, line 471: I just realised that if it wasn't for this string, `flate2` wouldn't need a release if `zlib-rs` increases its patch level.
Something I am wondering about is if this lock-step releasing is truly required for everything to work?

fintelia: As far as I can tell, the only thing that `zlib-rs` actually cares about is that [the first byte of the string is a '1' character](https://github.com/trifectatechfoundation/zlib-rs/blob/bd501f88a03aacf99163ba93431d870ae66865b5/libz-rs-sys/src/lib.rs#L1740).

PR #495, opened 2025-07-19
Correct documentation typo

fintelia:
As described [here](https://www.bolet.org/~pornin/deflate-flush.html) a partial flush involves writing one or two empty 10 _bit_ fixed code blocks.


Byron (approved):
Thanks a lot!

PR #496, opened 2025-07-19
Use partial flushes with miniz_oxide backend

fintelia:
I'll admit that I don't quite know why the code previously mapped `FlushCompress::Partial` to a sync flush

### DIFF
--- a/src/ffi/rust.rs
+++ b/src/ffi/rust.rs
@@ -136,7 +136,8 @@ impl From<FlushCompress> for MZFlush {
     fn from(value: FlushCompress) -> Self {
         match value {
             FlushCompress::None => Self::None,
-            FlushCompress::Partial | FlushCompress::Sync => Self::Sync,
+            FlushCompress::Partial => Self::Partial,
+            FlushCompress::Sync => Self::Sync,
             FlushCompress::Full => Self::Full,
             FlushCompress::Finish => Self::Finish,
         }


src/ffi/rust.rs
```diff
@@ -136,7 +136,8 @@ impl From<FlushCompress> for MZFlush {
     fn from(value: FlushCompress) -> Self {
         match value {
             FlushCompress::None => Self::None,
-            FlushCompress::Partial | FlushCompress::Sync => Self::Sync,
+            FlushCompress::Partial => Self::Partial,
```
Byron, line 139: @oyvindln Do you know if there was a reason *not* to use partial flushes before? If not, then it's probably safe to map it 1:1, but I wanted to check with you.
Thanks for your help!

Byron (commented):
Thanks a lot for looking into this!

I think it's best to validate this change, which to my mind *should* be good.

src/ffi/rust.rs
oyvindln: I have no idea

Byron: I looked into it by digging through the history, and it's unclear why things are the way they are.

There was a time when there was direct conversion.
```Rust
        flush: FlushCompress,
    ) -> Result<Status, CompressError> {
        let flush = MZFlush::new(flush as i32).unwrap();
```

The revision that actually introduced this line didn't mention why either.

So I think it's fair to merge this.

fintelia:
Yes, I discovered after this PR that `miniz_oxide` didn't support partial flushes, so I added them in https://github.com/Frommi/miniz_oxide/pull/179. Thus, `flate2` should avoid shipping this change until `miniz_oxide` makes a new release and this crate's version dependency is updated accordingly. With this change and the patch to `miniz_oxide`, #498 confirms that all backends implement all the flush types properly (at least for the one specific input that is tested).

(Someone should also probably make a PR to update the `1 | 2 => MZFlush::Sync` line in `miniz_oxide`, I missed it when making my PR)

fintelia:
Ugh, I probably should have reverted this PR until miniz_oxide made a new release. I think the current state of having this change in but using old miniz_oxide might result in calling "partial flush" not flushing at all.

PR #497, opened 2025-07-20
Undo introducing straight up incorrect documentation

Shnatsel:
Undo introducing misleading documentation, following the clarifications from miniz_oxide author in https://github.com/rust-lang/flate2-rs/pull/430#issuecomment-3091572081

I think it is still desirable to have some support for miniz_oxide level 10, but doing it properly will require code changes and not just documentation. For example, letting the API user set 10 and automatically lowering it to 9 when using one of the zlib backends, so that Cargo feature unification wouldn't outright break people's code. That will need some API design work, so I want to get the misleading docs out of the way ASAP. Sorry :sweat:

### DIFF
--- a/src/lib.rs
+++ b/src/lib.rs
@@ -190,18 +190,6 @@ impl Compression {
     ///
     /// The integer here is typically on a scale of 0-9 where 0 means "no
     /// compression" and 9 means "take as long as you'd like".
-    ///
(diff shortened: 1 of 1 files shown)


Byron (approved):
Thank you!

No worries about it - this happens, and in the end, it's me who should have done a better review as well.

PR #498, opened 2025-07-20
Upgrade miniz_oxide and add unit tests of the different flush kinds

fintelia:
This PR adds tests of the various flush kinds to ensure that all backends implement them consistently. The cases are carefully constructed so that even quite different backend implementations are likely to produce the expected output. In particular, it counts on the encoder detecting when the input is/isn't compressible and writing either compressed or stored blocks accordingly.  

Tests pass for me with the `zlib`, `zlib-ng`, `zlib-rs` and `cloudflare_zlib` backends. The `miniz_oxide` backend requires https://github.com/Frommi/miniz_oxide/pull/179 as otherwise it treats partial flushes as "no flush" (!)

### DIFF
--- a/src/ffi/miniz_oxide.rs
+++ b/src/ffi/miniz_oxide.rs
@@ -136,7 +136,8 @@ impl From<FlushCompress> for MZFlush {
     fn from(value: FlushCompress) -> Self {
         match value {
             FlushCompress::None => Self::None,
-            FlushCompress::Partial | FlushCompress::Sync => Self::Sync,
+            FlushCompress::Partial => Self::Partial,
+            FlushCompress::Sync => Self::Sync,
             FlushCompress::Full => Self::Full,
             FlushCompress::Finish => Self::Finish,
         }
--- a/src/mem.rs
+++ b/src/mem.rs
@@ -663,7 +663,6 @@ mod tests {
     use crate::write;
     use crate::{Compression, Decompress, FlushDecompress};
 
-    #[cfg(feature = "any_zlib")]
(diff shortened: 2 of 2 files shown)


Cargo.toml
```diff
@@ -112,3 +112,6 @@ any_impl = []
 [package.metadata.docs.rs]
 all-features = true
 rustdoc-args = ["--cfg", "docsrs"]
+
+[patch.crates-io]
+miniz_oxide = { git = "https://github.com/Frommi/miniz_oxide", branch = "master" }
```
fintelia: The only thing left should be getting a miniz_oxide 0.8.10 release, so this patch isn't necessary for the tests to pass

oyvindln:
This should be fine to add now, the miniz_oxide changes have been implemented.

Byron (approved):
Thanks a lot for your help with this, looks good to me!

fintelia:
Uh, that review might be for a different change. This PR doesn't touch any of those methods

Byron:
Uff, sorry for that. It can sometimes be very indirect, but when that happens it also explains the chain of reasoning which didn't happen here. So what really happened is that it compared to my local `main` which wasn't uptodate.
Yet I concluded that I just can't follow, but there was nothing to follow in the first place.

Here is the updated review for completeness, so nothing in the way of pressing the merge button.

----

The change corrects the miniz_oxide flush mapping and the added tests pass for the default backend as well as zlib and zlib-rs in this environment. I did not find any discrete regressions or compatibility issues introduced by this patch

PR #502, opened 2025-09-27
Add `(de)compress_uninit` that accepts `&[MaybeUninit<u8>]`

NobodyXu:
Having a safe API for `MaybeUninit<u8>` would be necessary for async-compression to wrap and reduce the performance penalty of initializing the buffer for all `tokio::io::*` traits implementation.

Also update `(de)compress_vec` API to use the new API instead of passing a `&mut [u8]` which technically violates rust's semantics.

NOTE that for `miniz_oxide` implementation, it still zero-initialize the `output`, since probably no one use it in production and now libz-rs is the go-to high-performance pure-rust implementation I don't see any point in optimizing it.

Related PR: trifectatechfoundation/bzip2-rs#147

### DIFF
--- a/src/ffi/mod.rs
+++ b/src/ffi/mod.rs
@@ -2,6 +2,15 @@
 
 use crate::mem::{CompressError, DecompressError, FlushCompress, FlushDecompress, Status};
 use crate::Compression;
+use std::mem::MaybeUninit;
+
+fn initialize_buffer(output: &mut [MaybeUninit<u8>]) -> &mut [u8] {
+    // SAFETY: Here we zero-initialize the output and cast it to [u8]
+    unsafe {
+        output.as_mut_ptr().write_bytes(0, output.len());
+        &mut *(output as *mut [MaybeUninit<u8>] as *mut [u8])
+    }
+}
 
 /// Traits specifying the interface of the backends.
 ///
@@ -20,6 +29,14 @@ pub trait InflateBackend: Backend {
         output: &mut [u8],
         flush: FlushDecompress,
     ) -> Result<Status, DecompressError>;
+    fn decompress_uninit(
+        &mut self,
+        input: &[u8],
+        output: &mut [MaybeUninit<u8>],
+        flush: FlushDecompress,
+    ) -> Result<Status, DecompressError> {
+        self.decompress(input, initialize_buffer(output), flush)
+    }
     fn reset(&mut self, zlib_header: bool);
 }
 
@@ -31,6 +48,14 @@ pub trait DeflateBackend: Backend {
         output: &mut [u8],
         flush: FlushCompress,
     ) -> Result<Status, CompressError>;
+    fn compress_uninit(
+        &mut self,
+        input: &[u8],
+        output: &mut [MaybeUninit<u8>],
+        flush: FlushCompress,
+    ) -> Result<Status, CompressError> {
+        self.compress(input, initialize_buffer(output), flush)
+    }
     fn reset(&mut self);
 }
 
--- a/src/mem.rs
+++ b/src/mem.rs
@@ -1,6 +1,7 @@
 use std::error::Error;
 use std::fmt;
 use std::io;
+use std::mem::MaybeUninit;
 
 use crate::ffi::{self, Backend, Deflate, DeflateBackend, ErrorMessage, Inflate, InflateBackend};
 use crate::Compression;
@@ -337,6 +338,20 @@ impl Compress {
         self.inner.compress(input, output, flush)
     }
 
+    /// Similar to [`Self::compress`] but accepts uninitialized buffer.
+    ///
+    /// If you want to avoid the overhead of zero initializing the
+    /// buffer and you don't want to use a [`Vec`], then please use
+    /// this API.
+    pub fn compress_uninit(
+        &mut self,
+        input: &[u8],
+        output: &mut [MaybeUninit<u8>],
+        flush: FlushCompress,
+    ) -> Result<Status, CompressError> {
+        self.inner.compress_uninit(input, output, flush)
+    }
+
     /// Compresses the input data into the extra space of the output, consuming
     /// only as much input as needed and writing as much output as possible.
     ///
@@ -351,12 +366,15 @@ impl Compress {
         output: &mut Vec<u8>,
         flush: FlushCompress,
     ) -> Result<Status, CompressError> {
-        write_to_spare_capacity_of_vec(output, |out| {
-            let before = self.total_out();
-            let ret = self.compress(input, out, flush);
-            let bytes_written = self.total_out() - before;
-            (bytes_written as usize, ret)
-        })
+        // SAFETY: bytes_written is the number of bytes written into `out`
+        unsafe {
+            write_to_spare_capacity_of_vec(output, |out| {
+                let before = self.total_out();
+                let ret = self.compress_uninit(input, out, flush);
+                let bytes_written = self.total_out() - before;
+                (bytes_written as usize, ret)
+            })
+        }
     }
 }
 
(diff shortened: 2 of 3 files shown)


src/mem.rs
```diff
@@ -582,18 +603,17 @@ impl fmt::Display for CompressError {
...
 ) -> T {
     let cap = output.capacity();
     let len = output.len();
 
-    output.resize(output.capacity(), 0);
-    let (bytes_written, ret) = writer(&mut output[len..]);
-
-    let new_len = core::cmp::min(len + bytes_written, cap); // Sanitizes `bytes_written`.
-    output.resize(new_len, 0 /* unused */);
+    unsafe {
```
Byron, line 611: It's probably not too common in this codebase, but should we start to explain why this is safe?

src/ffi/mod.rs
```diff
@@ -2,6 +2,14 @@
 
 use crate::mem::{CompressError, DecompressError, FlushCompress, FlushDecompress, Status};
 use crate::Compression;
+use std::mem::MaybeUninit;
+
+fn initialize_buffer(output: &mut [MaybeUninit<u8>]) -> &mut [u8] {
+    unsafe {
```
Byron, line 8: It's probably not too common in this codebase, but should we start to explain why this is safe?
The handling of MaybeUninit definitely gives me trouble, while it should be noted that I basically don't use `unsafe` so this looks like a different language to me 😅.

Byron (commented):
Thanks a lot for contributing! To me it seems very valuable to clean up output memory handling to maybe make it more robust thanks to the (to me very magical) `MaybeUninit`.

The biggest problem I am having is the MSRV failure, as I don't know why it happens now and what should be done about it. As any cloudflare-zlib or MSRV change should have triggered this back in the day, we may probably assume that the macOS image changed. 
Since cloudflare-zlib is on the way out, it might be acceptable to let it rot and disable this  particular test for it. But I am not certain the easy way is the good way here.

kornelski:
You're using cloudflare-zlib-sys 0.3.5, and the issue has been fixed in 0.3.6 released in March.

The incompatibility with new macOS headers was caused by an obsolete workaround meant for a very ancient MacOS, carried over from old zlib.h.

src/mem.rs
```diff
@@ -580,18 +607,20 @@ impl fmt::Display for CompressError {
...
     let cap = output.capacity();
     let len = output.len();
 
-    output.resize(output.capacity(), 0);
-    let (bytes_written, ret) = writer(&mut output[len..]);
-
-    let new_len = core::cmp::min(len + bytes_written, cap); // Sanitizes `bytes_written`.
-    output.resize(new_len, 0 /* unused */);
+    let (bytes_written, ret) = writer(output.spare_capacity_mut());
+    output.set_len(cap.min(len + bytes_written)); // Sanitizes `bytes_written`.
```
jongiddy, line 623: I'd say this should panic if `len + bytes_written > cap` but the non-panicking min behavior is that same as the existing code.

jongiddy (approved):

src/mem.rs
NobodyXu: Agree that panicking is probably the best behavior, though I don't want to change the API of the _vec function

Byron: Agreed, let's do this in a follow-up, and also, let's do it so that the calling context can produce an error instead.

src/mem.rs
```diff
@@ -455,6 +469,16 @@ impl Decompress {
         self.inner.decompress(input, output, flush)
     }
 
+    /// Similar to [`Self::decompress`] but accepts uninitialized buffer
```
Byron, line 472: If we are adding a new public method, I think it should also be made clear why we'd use one over the other.
Alternatively, we can make it private.

src/mem.rs
```diff
@@ -337,6 +338,16 @@ impl Compress {
         self.inner.compress(input, output, flush)
     }
 
+    /// Similar to [`Self::compress`] but accepts uninitialized buffer.
```
Byron, line 341: If we are adding a new public method, I think it should also be made clear why we'd use one over the other.
Alternatively, we can make it private.

Byron (commented):
Thanks a lot for everyone's help!

I take it that this PR can be merged as is then, even though I have a few more coment.

Could you also rebase and squash it down a bit?

src/mem.rs
NobodyXu: Thanks, updated the doc to explain why they should use that

src/mem.rs
NobodyXu: Thanks, updated the doc to explain why they should use that

Byron (approved):
Thank a lot!

Then… let's do it 🙏.

*(I am always a bit afraid to merge anything here with my possibly dangerous mix of naiveté and ignorance 😅)*

Byron:
I had no plans, actually. Maybe we can wait a little longer

```
2025-09-29 18:28 +0200 Sebastian Thiel            M─┐ [main] {origin/main} {origin/HEAD} Merge pull request #502 from NobodyXu/patch-1
2025-09-29 10:52 +0000 Jiahao XU                  │ o Add `(de)compress_uninit` impl for uninit buffer
2025-09-29 04:37 +0200 Sebastian Thiel            M─┤ Merge pull request #503 from jongiddy/update-cloudflare-zlib
2025-09-28 21:21 +0100 Jonathan Giddy             │ o Update cloudflare-zlib-sys crate
2025-07-20 09:20 +0200 Sebastian Thiel            M─┤ Merge pull request #497 from Shnatsel/undo-misleading-docs
2025-07-20 01:09 +0100 Sergey "Shnatsel" Davidoff │ o Undo introducing straight up incorrect documentation
2025-07-19 14:43 +0200 Sebastian Thiel            M─┤ Merge pull request #496 from fintelia/flush-kinds
2025-07-19 00:43 -0700 Jonathan Behrens           │ o Use partial flushes with miniz_oxide backend
2025-07-19 10:22 +0200 Sebastian Thiel            M─│─┐ Merge pull request #495 from fintelia/patch-1
2025-07-19 00:18 -0700 Jonathan Behrens           │ │ o Correct documentation typo
2025-07-03 07:37 +0200 Sebastian Thiel            M─┼─┘ Merge pull request #492 from Shnatsel/simd-adler32
2025-07-03 07:30 +0200 Sebastian Thiel            │ o [simd-adler32] [refs/patches/simd-adler32/thanks-clippy] thanks clippy
2025-07-02 14:08 +0100 Sergey "Shnatsel" Davidoff │ o Switch from adler2 to simd-adler32 crate when using miniz_oxide backend
2025-06-14 14:24 +0200 Sebastian Thiel            M─┤ Merge pull request #491 from folkertdev/zlib-version
2025-06-07 12:48 +0200 Folkert de Vries           │ o use `zlibVersion()` instead of a `const` for the version
2025-06-07 05:21 +0200 Sebastian Thiel            M─┤ <1.1.2> Merge pull request #490 from folkertdev/update-zlib-rs-0.5.1
```

There doesn't seem to be too much either.
What do you recommend me doing?

NobodyXu:
Looks like it has some fixes (use `zlibVersion()` instead), and I think it's actually good to have too much commits accumulating?

cc cuts a new release every week as long as there're changes, not saying flate2 should be as frequent as that, but I think letting commits sit for months without release is actually bad as end users end up waiting really long for improvements in flate2.

plus, having less commits per release is good for finding bugs, users who experienced breakage after upgrading can find the culprit commit easier than having a lot.

With that says it's my two cents, and I'm not rushed to have this PR released, just my own opinion on release cadence

Byron:
Thanks for sharing! Let me try to get better at making releases then :).

Oh, and I remember now, I can't approve my own PRs so releasing is a hassle unless it's planned and a version bump is made by someone else. I don't actually know who is associated with this project and hope that everyone who is suggested as reviewer also is allowed to approve PRs here.
Please see https://github.com/rust-lang/flate2-rs/pull/504 .

NobodyXu:
Thank you!

NobodyXu:
The docs.rs build fails brcause doc_auto_cfg has been removed in nightly

https://docs.rs/crate/flate2/1.1.3/builds/2560338

Byron:
Thanks for reporting! If it was an issue I'd add the"help wanted" label just because I can't merge my own fixes 😅

mstange:
I think this PR may have forgotten the equivalent changes to the Inflate and Deflate implementations in src/ffi/zlib_rs.rs, is that possible? I've filed #544 about it.

PR #505, opened 2025-10-05
Fix docs.rs build

NobodyXu:
Nightly feature doc_auto_cfg is merged into doc_cfg rust-lang/rust#138907


Byron (approved):
Thanks a lot!

If you think the documentation changes are worth another release, you could increment the patch level as well.

NobodyXu:
thanks, opened #506

Issue #508, opened 2025-10-23
Missing last two Bytes when decompressing


dns2utf8:
Hi all

I have a weird decompression bug with deflate. I isolated it to this buffer:

```rust
let hex_buffer = [0xaa, 0x56, 0x72, 0x4b, 0x2d, 0x49, 0xce, 0xf0, 0xc8, 0xcf, 0xcf, 0xf6, 0x4d, 0x2d, 0x49, 0x4c, 0x49, 0x2c, 0x49, 0x54, 0xb2, 0xaa, 0x56, 0xca, 0xc8, 0xcf, 0xcf, 0x56, 0xb2, 0x02, 0x53, 0xc5, 0xfa, 0x19, 0xa9, 0x39, 0x39, 0xf9, 0x4a, 0x3a, 0x4a, 0xd9, 0xa9, 0x95, 0x4a, 0x56, 0x4a, 0x89, 0x79, 0x95, 0x19, 0xf9, 0xe5, 0x4a, 0x00, 0x00, 0xff, 0xff];

  let mut decoder = read::DeflateDecoder::new(&*hex_buffer);

            let mut buf = Vec::new();
            let decoded_len = decoder
                .read_to_end(&mut buf)
                .context("unable to read compressed payload")?;

let s = String::from_utf8(buf).context("utf-8 error")?;
let expected =  "{\"FetchHookMetadata\":{\"hook\":\"hooks/hello\",\"key\":\"anyhow\"}}";
```

This should return this String, but it is missing the last two `}}`

Since this buffer will be coming over WebSocket I extended it with `[ 0x00, 0x00, 0xff, 0xff ]` but not doing that made no difference.

Hope anyone has an idea on why it just does not decompress the last two Bytes.
Thank you in advance,
Stefan

fintelia:
How are you producing the compressed data? It doesn't actually include those final two bytes. Which either means the compressor wasn't flushed when writing the output, or if you're using this library, the [recent flush changes](https://github.com/rust-lang/flate2-rs/pull/496) might have broken it.

Byron:
I don't know what I was thinking, given that the PR is still in draft and failing CI - it's not even merged in yet.
It should have been clear to me that that the PR here depends on the one in `miniz_oxide`.

Oh well, @fintelia can you come up with a PR that reverts the flush changes and bumps the patch level of this crate? Then I can merge it and create a new release. I'd do it myself, but can't self-merge here.

Thanks a million.

[...]

PR #509, opened 2025-10-24
Revert flush change

fintelia:
See https://github.com/rust-lang/flate2-rs/issues/508#issuecomment-3440498105 for context


Byron (approved):
Thanks so much! A new release is on the way.

PR #510, opened 2025-10-24
chore: make some documents clearer

reddaisyy:
make some documents clearer


Byron (approved):
Thanks a lot!

Personally, I also like punctuation a lot, maybe there are more doc-comments that lack it for no good reason.

Issue #511, opened 2025-10-29
Docs don't explain how to choose between `bufread`/`read`/`write`


Zalathar:
The encoders and decoders in this crate are available in three variants: `bufread`, `read`, and `write`. This makes it necessary for users to choose a variant before they can compress or decompress anything.

However, the crate documentation gives almost no indication of what the practical differences are, or why someone would choose one over the others.

- Is one of them a sharp-edged performance hazard?
- Is one of them a “safe” default choice, at the risk of being sub-optimal?
- Given that encoding/decoding is orthogonal to reading/writing, what is the intended use case for each combination?
- What's best for the simple cases of decoding an entire file into memory, or encoding a memory buffer into a file?
  - See also #201

Byron:
Thanks a lot for bringing this up!

Even though I never tried to read the API documentation, I share the feeling of being somewhat confused by all these variants. It is my feeling that just a little bit of clarification would be enough to know how to think about these, with examples to hammer it home.

Is this a label you see?

<img width="368" height="88" alt="Image" src="https://github.com/user-attachments/assets/91ee2a93-ca52-4f9f-9139-f5cad5b6c22e" />

Maybe it's possible to figure this out with LLM help and then contribute the documentation changes you would have wanted before writing this issue?

Otherwise, it might end up like #201 - this codebase is entirely community maintained.
Thanks for your consideration.

Zalathar:
From a quick look at the code, it seems that the `bufread` implementations are the “real” reader-based implementations, and the `read` variants mostly just wrap their input stream in a `BufReader` and then delegate to the `bufread` implementation.

Zalathar:
Ah, b91748a also seems relevant.

It looks like the `read` types were originally the “real” ones, but they always added an unconditional layer of buffering in front of the underlying stream.

That change introduced variants that rely on the underlying stream already being buffered, and changed the `read` variants to be wrappers that explicitly add a buffering layer.

It's a bit strange to me that that commit introduces a bespoke `BufReader` type (which still exists today), instead of just using `std::io::BufReader`, which would have already existed in stable at that time.

Zalathar:
To make life easier for users, I wonder if it's better to *only* mention `bufread` and `write` in the main explanation/table.

Then `read` would only be mentioned in a relatively-minor note, as something that exists for backwards-compatibility with code that predates `bufread`.

[...]

PR #512, opened 2025-11-02
Explain rationale for choosing bufread,read,write modules

jongiddy:
Provide a clearer explanation of when to choose the different module implementations.

Make the `write` module example more specific to a problem that can only be handled using the `write` module.

Update inter-module links to [RFC-1946 format](https://rust-lang.github.io/rfcs/1946-intra-rustdoc-links.html).

Fixes https://github.com/rust-lang/flate2-rs/issues/511


Byron (approved):
Thanks a lot - this looks like an improvement that does fix #511.

While I am here, I am merging this PR but invite @Zalathar to follow it up with further improvements in case it's not entirely fitting just yet.

PR #513, opened 2025-11-11
use `zlib_rs` api

folkertdev:
This is an initial design for using the `zlib_rs` API, calling the safe functions that it exposes rather than the unsafe wrappers of `libz-rs-sys`.

I'm designing the flate2 side here, the actual implementation still uses `libz-rs-sys` because `zlib-rs` does not export some things that we need here.

Inspired by https://github.com/GitoxideLabs/gitoxide/pull/2155.

### DIFF
--- a/src/ffi/miniz_oxide.rs
+++ b/src/ffi/miniz_oxide.rs
@@ -145,8 +145,9 @@ impl From<FlushCompress> for MZFlush {
 
 impl DeflateBackend for Deflate {
     fn make(level: Compression, zlib_header: bool, _window_bits: u8) -> Self {
-        // Check in case the integer value changes at some point.
-        debug_assert!(level.level() <= 9);
+        // Check in case the integer value changes at some point. Unlike the other zlib
+        // implementations, miniz_oxide actually has a compression level 10.
+        debug_assert!(level.level() <= 10);
 
         let mut inner: Box<CompressorOxide> = Box::default();
         let format = format_from_bool(zlib_header);


src/ffi/zlib_rs.rs
```diff
@@ -0,0 +1,164 @@
...
+            FlushDecompress::Sync => Self::SyncFlush,
+            FlushDecompress::Finish => Self::Finish,
+        }
+    }
+}
+
+impl InflateBackend for Inflate {
+    fn make(zlib_header: bool, window_bits: u8) -> Self {
+        Inflate {
+            inner: ::zlib_rs::Inflate::new(zlib_header, window_bits),
```
Byron: I wonder if there are any benefits to have a pure-Rust implementation of this. Maybe it's less allocations?
In any case, I suppose it will have to allocate internal buffers.
This is just me writing down a minor thought, it's not really relevant.

src/ffi/zlib_rs.rs
```diff
@@ -0,0 +1,164 @@
+//! Implementation for `zlib_rs` rust backend.
```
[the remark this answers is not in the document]
folkertdev: I wrote some things but yeah generally it's just the most bare-bones thing that satisfies the assumptions that are made about a backend.

Byron (commented):
Absolutely fantastic, and thanks for these wonderful docs 🙏.

Byron:
Oh wow, I was close to merging this.
But we don't as I wait for your cue when `zlib-rs` is re-released.

Let's put the update in a separate commit as well please, it's easier to review if we stop rewriting now.

folkertdev:
We've released 0.5.3, containing the stable API. This is ready to merge now, and if there are no issues we can make it the default implementation.

folkertdev:
> Let's put the update in a separate commit as well please, it's easier to review if we stop rewriting now.

I missed that. What do you mean with the update? just the `Cargo.toml` changes?

src/ffi/miniz_oxide.rs
```diff
@@ -146,7 +146,7 @@ impl From<FlushCompress> for MZFlush {
 impl DeflateBackend for Deflate {
     fn make(level: Compression, zlib_header: bool, _window_bits: u8) -> Self {
         // Check in case the integer value changes at some point.
-        debug_assert!(level.level() <= 10);
+        debug_assert!(level.level() <= 9);
```
jongiddy: Why is this change necessary? `miniz_oxide` supports level 10 according to its [docs](https://docs.rs/miniz_oxide/latest/miniz_oxide/deflate/enum.CompressionLevel.html).

PR #514, opened 2025-11-29
use Decompress::reset() instead of recreating in DeflateDecoder

amirshukayev:
From https://www.zlib.net/manual.html

> This function is equivalent to inflateEnd followed by inflateInit, but does not free and reallocate the internal decompression state. The The stream will keep attributes that may have been set by inflateInit2. total_in, total_out, adler, and msg are initialized.

This led to a ~14% performance improvement on my workloads with `MultiGZDecoder`.

### DIFF
--- a/src/deflate/bufread.rs
+++ b/src/deflate/bufread.rs
@@ -166,7 +166,7 @@ pub struct DeflateDecoder<R> {
 }
 
 pub fn reset_decoder_data<R>(zlib: &mut DeflateDecoder<R>) {
-    zlib.data = Decompress::new(false);
+    zlib.data.reset(false);
 }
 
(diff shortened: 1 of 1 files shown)


jongiddy (approved):
Yes, it would be nice to know why this was changed many years ago to avoid the reset. I cannot see a problem with accepting it now.

Byron:
Thank you! Then I'd say we merge it, as I'd expect to have seen 'something codified' coming with `Decompress::new()` that prevents future selfs from changing it back.

In the worst case we have to roll it back, which seems particularly easy in this case, in the best there is a significant performance improvement for those who have been using the `MultiGZDecoder`.

Issue #517, opened 2025-12-07
1.1.7 have breaking change about feature `zlib-rs`


youknowone:
```diff
-zlib-rs = ["any_zlib", "libz-rs-sys"]
+zlib-rs = ["any_impl", "dep:zlib-rs"]
```

`zlib-rs` doesn't enable `any_zlib` anymore, so APIs under `any_zlib` like `Compress::new_gzip`, `Compress::set_dictionary` became unexisting.

introduced by 5b24d983 (#513)

I assume this might be intended, but it is surprising change between 1.1.5 -> 1.1.7 because I can't pin the version as `1.1`

What will be the migration guide in that case? 

related: #515

markusmoenig:
Great job guys introducing breaking changes at 1.1.7 !!!

Byron:
Thanks a lot for sharing!

This certainly means that neither me nor the other reviewer nor the author fully understand our feature toggles.
It also means that CI isn't able to detect this kind of breakage, which is understandable if it uses the same toggles to turn on or off certain tests.

Let's start with amping up the documentation around `any_zlib` and `any_impl`, and see if we can get towards a fix from there.

Meanwhile, the latest release was yanked.

youknowone:
Thank you so much! is there anything I can help?

Byron:
Thanks for offering, but I don't think so. `zlib-rs` without C bindings is back to the drawing board for now, and a new release can be made once these are sorted.

PR #521, opened 2025-12-07
Allow zlib_impl for more functions

youknowone:
ref: #517

At least these functions are zlib-impl compatible.

I guess tests like `set_dictionary_raw` also expected to be marked to `zlib-impl` to guarantee compatibility


youknowone:
Oh, #520 already did it.

PR #526, opened 2026-01-13
use `zlib-rs` for crc32 (when available)

folkertdev:
This currently hits a bug in the zlib-rs implementation that we don't run into internally, but that the public API does. So, I'll need to go fix that. But hopefully this gives an impression what it'll look like.


folkertdev:
With that this should be OK to merge. We _might_ change the path of where the crc are exported at some point in the future, but that should be straightforward to handle when bumping the dependency version in flate2.

Byron (approved):
Thanks  a lot, I think that's it!

PR #527, opened 2026-01-13
Add tests to check data CRC

jongiddy:
The code currently has no tests for the data CRC. Add a test that the GzEncoders generate the CRC expected by RFC 1952 and that the GzDecoders fail if the CRC does not match.

This will give us more confidence to make any changes to the CRC calculation (e.g. #523 or #526).

### DIFF
--- a/src/gz/mod.rs
+++ b/src/gz/mod.rs
@@ -573,6 +573,165 @@ mod tests {
     }
 
     #[test]
+    fn gzip_encoder_matches_rfc1952() {
+        /// Extract CRC32 and ISIZE from gzip footer
+        fn extract_zip_footer(compressed: &[u8]) -> (u32, u32) {
+            assert!(compressed.len() >= 8, "Gzip output too short");
+            let footer_start = compressed.len() - 8;
+
+            let crc = u32::from_le_bytes([
+                compressed[footer_start],
+                compressed[footer_start + 1],
+                compressed[footer_start + 2],
+                compressed[footer_start + 3],
+            ]);
+
+            let size = u32::from_le_bytes([
+                compressed[footer_start + 4],
+                compressed[footer_start + 5],
+                compressed[footer_start + 6],
+                compressed[footer_start + 7],
+            ]);
+
+            (crc, size)
+        }
+
+        #[track_caller]
+        fn test_crc_for_write(data: &[u8], expected_crc: u32, description: &str) {
+            // Compress data using write::GzEncoder
+            let mut encoder = write::GzEncoder::new(Vec::new(), Compression::default());
+            encoder.write_all(data).unwrap();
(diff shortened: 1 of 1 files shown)


Byron (approved):
Thanks a lot, I also think these will be very useful!

I'd wave the PR through, but then again, I have no clue. Maybe @folkertdev could take a look to see what he thinks might be missing. Based on this, we can decide if there should be more, or of that 'more' can or should be in another PR.
But maybe there are legitimate change requests as well, so I'd want to wait for his opinion.

src/gz/mod.rs
```diff
@@ -572,6 +572,162 @@ mod tests {
...
+            let mut encoder = write::GzEncoder::new(Vec::new(), Compression::default());
+            encoder.write_all(data).unwrap();
+            let compressed = encoder.finish().unwrap();
+
+            let expected_size = data.len() as u32;
+            let (actual_crc, actual_size) = extract_zip_footer(&compressed);
+
+            assert_eq!(
+                expected_crc, actual_crc,
+                "CRC32 mismatch for write {}: expected 0x{:08x}, got 0x{:08x}",
```
folkertdev:
```suggestion
                "CRC32 mismatch for write {}: expected {:#08x}, got {:#08x}",
```

not important but sort of a neat trick you might not have encountered.

src/gz/mod.rs
```diff
@@ -572,6 +572,162 @@ mod tests {
...
+                compressed[footer_start + 4],
+                compressed[footer_start + 5],
+                compressed[footer_start + 6],
+                compressed[footer_start + 7],
+            ]);
+
+            (crc, size)
+        }
+
+        fn test_crc_for_write(data: &[u8], expected_crc: u32, description: &str) {
```
folkertdev:
```suggestion
        #[track_caller]
        fn test_crc_for_write(data: &[u8], expected_crc: u32, description: &str) {
```

folkertdev (commented):
Some minor style ideas/suggestions. Overall I think this gives good coverage of how `flate2` uses checksums.

my PR additionally tests the public API (e.g. `combine` and `reset`). So together I think that would cover crc usage well.

Byron:
Thanks a lot for chiming in @folkertdev! @jongiddy, I will wait until you had a chance to look at the suggestions, but generally will be happy to merge.

folkertdev (approved):

Byron (commented):
Thanks a lot everyone! I'd merge this.

However, please allow one last question: could these tests be in `tests/` as they don't use private APIs? It looks like it, and I would have checked (and possibly done) the move myself but then I can't merge it anymore 😅.

Then again, as this doesn't introduce tests, we could also move the tests over, if possible, in a separate PR and call this one done enough.

PR #529, opened 2026-01-23
upgrade zlib-rs to version `0.6.0`

folkertdev:
https://github.com/trifectatechfoundation/zlib-rs/releases/tag/v0.6.0

Not that eventful from the flate2 perspective. When compiling with avx512 compression should be a bit faster.


folkertdev:
Can we have a release with this change (the PR already bumps the version number). Because it's a "major" change it won't be upgraded to automatically.

PR #547, opened 2026-05-02
Fix infinite loop in writing gzip header/footer

Shnatsel:
## Summary

Fix `write::GzEncoder` spinning forever when the wrapped writer returns `Ok(0)` while writing gzip header or footer bytes.

## Problem

Most write paths in flate2 go through `zio::Writer::dump`, which treats `Ok(0)` for a non-empty buffer as `io::ErrorKind::WriteZero`.

`write::GzEncoder` has two gzip-specific direct write loops:
- writing the gzip header
- writing the final 8-byte gzip footer

Those loops previously retried after `Ok(0)` without making progress. A writer that repeatedly returns `Ok(0)` for a non-empty buffer could therefore cause `GzEncoder::try_finish`/`finish`/drop finalization to spin indefinitely.

## Fix

Add a small helper for gzip’s direct writes that mirrors the shared writer behavior: `Ok(0)` on a non-empty buffer is converted to `WriteZero`.

## Tests

Added regression coverage for both spin sites:
- zero-length progress while writing the gzip header
- zero-length progress while writing the gzip footer

Validation:
- `cargo test --test zero-write`
- `cargo test gz::tests`
- `cargo test`

## Credits

The issue was discovered by GPT-5.5 xhigh

### DIFF
--- a/tests/zero-write.rs
+++ b/tests/zero-write.rs
@@ -92,6 +92,10 @@ fn gzip_header_zero_write_is_error() {
     // so the loop retried the same write forever. This bounded writer turns
     // that spin into an ordinary error so the test can catch the missing
     // WriteZero handling without hanging.
+    //
+    // Three zero-length writes is arbitrary but enough to show the old code
+    // was retrying instead of treating Ok(0) on a non-empty buffer as
+    // WriteZero.
     let writer = AlwaysZeroThenError::new(3);
     let mut encoder = flate2::write::GzEncoder::new(writer, flate2::Compression::default());
 
@@ -105,6 +109,9 @@ fn gzip_footer_zero_write_is_error() {
     // writer that accepts the header and deflate payload but returns Ok(0) for
     // the empty-stream footer (CRC32 0, ISIZE 0) left crc_bytes_written
     // unchanged, so try_finish retried the same footer slice forever.
+    //
+    // Three zero-length footer writes is arbitrary but enough to show the old
+    // code was retrying the same footer slice instead of returning WriteZero.
     let writer = ZeroOnGzipFooterThenError::new(3);
     let mut encoder = flate2::write::GzEncoder::new(writer, flate2::Compression::default());
 
--- a/tests/zero-write.rs
+++ b/tests/zero-write.rs
@@ -65,7 +65,7 @@ impl Write for ZeroOnGzipFooterThenError {
             return Ok(0);
         }
 
-        if buf.len() == 8 {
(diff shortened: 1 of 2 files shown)


tests/zero-write.rs
```diff
@@ -1,6 +1,113 @@
...
+}
+
+#[test]
+fn gzip_header_zero_write_is_error() {
+    // GzEncoder used to spin in write_header when the wrapped writer returned
+    // Ok(0) for the non-empty gzip header buffer: the header was not drained,
+    // so the loop retried the same write forever. This bounded writer turns
+    // that spin into an ordinary error so the test can catch the missing
+    // WriteZero handling without hanging.
+    let writer = AlwaysZeroThenError::new(3);
```
Byron, line 99: Why `(3)`? Maybe put it into a variable and comment, or make it non-configurable and document it at use.

tests/zero-write.rs
```diff
@@ -1,6 +1,113 @@
...
+    assert_eq!(err.kind(), io::ErrorKind::WriteZero);
+}
+
+#[test]
+fn gzip_footer_zero_write_is_error() {
+    // GzEncoder also used to spin while writing the 8-byte gzip footer. A
+    // writer that accepts the header and deflate payload but returns Ok(0) for
+    // the footer left crc_bytes_written unchanged, so try_finish retried the
+    // same footer slice forever.
+    let writer = ZeroOnGzipFooterThenError::new(3);
```
Byron, line 115: Why `(3)`? Maybe put it into a variable and comment, or make it non-configurable and document it at use.

Byron (changes requested):
Thanks a lot for the fix!

This looks good to me, just a few comments I'd like to add to make it clear where some magic numbers are coming from.

tests/zero-write.rs
```diff
@@ -1,6 +1,113 @@
...
+                Err(io::Error::new(
+                    io::ErrorKind::Other,
+                    "gzip encoder retried a zero-length footer write",
+                ))
+            } else {
+                Ok(0)
+            }
+        } else {
+            Ok(buf.len())
+        }
```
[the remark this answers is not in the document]
Byron: I think we can hardcode this, possibly add a note that says what to do if this fails (or why this always works).

Shnatsel:
Thanks for the quick review! Both your and copilot comments should now be addressed.

Byron (approved):
Great, let's merge this.

PR #550, opened 2026-06-18
test: generate corrupt gzip fixture at runtime

### LINKED ISSUE #549: Test for corrupt gz file trips up corporate security
Hello flate2 maintainers!

My organization's security team had a minor fit over the corrupt-gz-file.bin file when I was trying to pull sources for building Firefox. I eventually convinced them that corrupt-gz-file.bin in the tests directory is benign and intentionally corrupt/nonsensical data, but it was pointed out to me that the file could theoretically be a malicious encrypted payload waiting to be extracted and executed.

Personally I don't think this is the case... but it had me thinking whether or not the file could be effectively generated at test runtime with random contents to satisfy the test coverage **and** my security team. If that's an acceptable approach, I'm happy to attempt creating a PR for submission; I don't use rust very often but I think I could figure it out if the approach doesn't seem crazy.

nanookclaw:
## Summary

Closes #549.

This removes `tests/corrupt-gz-file.bin` and updates `test_extract_failure` to generate the corrupt gzip input in memory. The test now writes a small valid gzip stream with `GzEncoder`, flips a byte in the CRC32 portion of the gzip footer, then verifies `GzDecoder` still reports `io::ErrorKind::InvalidInput`.

That keeps the same failure-path coverage without shipping an opaque gzip-looking binary fixture that can trip security scanners.

## Verification

- `cargo fmt --check`
- `cargo test --test gunzip`
- `cargo test --test gunzip --features zlib-rs --no-default-features`
- `cargo test --test gunzip --features zlib --no-default-features`


EhhChris:
Beat me to it on a travel day, thank you!

tests/gunzip.rs
```diff
@@ -31,7 +33,15 @@ fn test_extract_success_partial_multi() {
 // test extraction fails on a corrupt file
 #[test]
 fn test_extract_failure() {
-    let result = extract_file(Path::new("tests/corrupt-gz-file.bin"));
+    let mut encoder = GzEncoder::new(Vec::new(), Compression::default());
+    encoder.write_all(b"corrupt me").unwrap();
```
Byron, line 37: 😁

Byron (approved):
Wonderful, thanks so much!

PR #553, opened 2026-07-22
Reject oversized gzip extra fields

### LINKED ISSUE #552: `GzBuilder::extra` silently creates invalid gzip streams when extra field length exceeds u16::MAX
### **Description**:

`GzBuilder::extra` accepts an arbitrary `Vec<u8>` for the gzip extra field. However, the gzip header stores the extra field length as a 16-bit little-endian integer. The encoder writes the length with a lossy cast:

```
header.extend((v.len() as u16).to_le_bytes());
header.extend(v);
```

As a result, when the provided extra field is longer than `u16::MAX`, the encoded length wraps around while the full extra data is still appended to the header. This produces a malformed gzip stream using only safe public APIs.

For example, an extra field of length `65536` is encoded with XLEN = `0`, but 65536 extra bytes are still written. A decoder then treats those extra bytes as compressed deflate data and fails.

### **Environment**:

flate2 version: 1.1.9

rustc version: 1.94.1

OS: Ubuntu 20.04.6 LTS

### PoC 

```rust
use std::io::{Read, Write};
use flate2::read::GzDecoder;
use flate2::{Compression, GzBuilder};

fn main() {
    let mut encoder = GzBuilder::new()
        .extra(vec![b'A'; 65536])
        .write(Vec::new(), Compression::default());

    encoder.write_all(b"hello").unwrap();
    let compressed = encoder.finish().unwrap();

    let mut decoded = Vec::new();
    let err = GzDecoder::new(&compressed[..])
        .read_to_end(&mut decoded)
        .expect_err("gzip produced by GzBuilder with oversized extra should not round-trip");

    eprintln!("decoder error: {err}");
}
```

### **Expected behavior:**

One of the following should happen:

1. `GzBuilder` should reject extra fields longer than `u16::MAX`.
2. The API should return an error when attempting to build/finish such a gzip stream.
3. The method should panic with a documented message, similar to how `filename` and `comment` document panic behavior for invalid input.

### **Actual behavior**

```
decoder error: corrupt deflate stream
```
### **Root cause**:

The issue appears to be in `GzBuilder::into_header`:

https://github.com/rust-lang/flate2-rs/blob/167053740e4e73d64bc644bebedcc11718cc2f4c/src/gz/mod.rs#L403-L407

The cast from `usize` to `u16` truncates the length. For `v.len() == 65536`, the encoded length becomes `0`, while the full extra buffer is still appended.

### **Possible Fix**

A minimal fix would be to validate the extra field length before writing the header:

```
if let Some(v) = extra {
    assert!(
        v.len() <= u16::MAX as usize,
        "gzip extra field length must be <= u16::MAX"
    );
    flg |= FEXTRA;
    header.extend((v.len() as u16).to_le_bytes());
    header.extend(v);
}
```

Byron:
## Tasks

This section is for Byron only. Models continuing this PR must not add, remove, check, uncheck, rename, or reorder checkboxes here.

- [x] refackiew

----

Everything below this line was generated by `Codex GPT-5`.

Created by Codex on behalf of Byron. Byron will review before this is ready to merge.

Fixes #552

## Summary

Reject gzip extra fields larger than `u16::MAX` when they are supplied to `GzBuilder::extra`. This prevents XLEN from wrapping while the full extra field is serialized and matches the documented eager panic behavior of the filename and comment setters.

## Validation

- `cargo fmt -- --check`
- `cargo test gz::tests::extra_too_long`
- `cargo test gz::tests::fields`
- `cargo test --lib` (62 passed)

Codex commit review found no issues. Its additional `cargo test --all-features` attempt could not download `libz-ng-sys` because DNS resolution for crates.io failed.

### DIFF
--- a/src/gz/mod.rs
+++ b/src/gz/mod.rs
@@ -1,3 +1,4 @@
+use std::convert::TryFrom;
 use std::ffi::CString;
 use std::io::{BufRead, Error, ErrorKind, Read, Result, Write};
 use std::time;
@@ -411,7 +412,12 @@ impl GzBuilder {
         let mut header = vec![0u8; 10];
         if let Some(v) = extra {
             flg |= FEXTRA;
-            header.extend((v.len() as u16).to_le_bytes());
+            header.extend(
+                (u16::try_from(v.len()).expect(
+                    "`extra` can only be created from `extra()` which would have panicked on len > u16::MAX",
+                ))
+                .to_le_bytes(),
+            );
             header.extend(v);
         }
         if let Some(filename) = filename {
--- a/src/gz/mod.rs
+++ b/src/gz/mod.rs
@@ -338,8 +338,17 @@ impl GzBuilder {
     }
 
     /// Configure the `extra` field in the gzip header.
+    ///
+    /// # Panics
+    ///
+    /// Panics if `extra` is longer than [`u16::MAX`].
     pub fn extra<T: Into<Vec<u8>>>(mut self, extra: T) -> GzBuilder {
-        self.extra = Some(extra.into());
+        let extra = extra.into();
+        assert!(
+            extra.len() <= u16::MAX as usize,
+            "gzip extra field length cannot exceed u16::MAX"
+        );
+        self.extra = Some(extra);
         self
     }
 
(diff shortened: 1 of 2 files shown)


src/gz/mod.rs
jongiddy: This line could change to use:
```rust
            header.extend((u16::try_from(v.len()).unwrap()).to_le_bytes());
```

It is redundant right now, since your other change restricts `self.extra` to 16 bits, but would prevent creating invalid gzip if any other way to add `self.extra` was added later.

jongiddy (approved):

src/gz/mod.rs
Byron: A valid point! Doppelt hält besser!

This is implemented in https://github.com/rust-lang/flate2-rs/pull/553/commits/e3b8f5cb73f8430697eb022422ed5e94dcd88fa9 .

Byron:
Thanks for the timely review @jongiddy, it's much appreciated 🎉. I have implemented the requested change and will merge soon.

PR #556, opened 2026-07-27
Reject incomplete deflate streams at EOF

### LINKED ISSUE #499: Regression in decompressing bad data using miniz_oxide>=0.8.5 w/ small input
Copying the issue from miniz_oxide: https://github.com/Frommi/miniz_oxide/issues/174

xref: https://github.com/milesgranger/cramjam/issues/211

Seems a decompression error only arrives if the bad input data is over 4 characters long:

This does not raise an error:

```rust
    fn decompress_bad_data() {
        let mut decoder = DeflateDecoder::new(b"1".as_slice());
        let mut out = vec![];
        assert!(std::io::copy(&mut decoder, &mut out).is_err());
    }
```

But this will, 5 bytes at least for me, seems to spark the error, up to 4 bytes will not raise an error.

```rust
    fn decompress_bad_data() {
        let mut decoder = DeflateDecoder::new(b"12345".as_slice());
        let mut out = vec![];
        assert!(std::io::copy(&mut decoder, &mut out).is_err());
    }
```

---

And as pointed out in the linked issue, it actually depends on what the first byte is on whether it fails as expected or not - further adding to the awkwardness.

Guflly:
Read decoders could return EOF successfully when the input ended before the deflate stream reached StreamEnd.

Treat a no-progress Ok or BufError at input EOF as UnexpectedEof, and cover the short invalid inputs from #499 across the read decoder.

Fixes #499

### DIFF
--- a/src/deflate/mod.rs
+++ b/src/deflate/mod.rs
@@ -166,7 +166,9 @@ mod tests {
     fn rejects_incomplete_stream() {
         for input in [b"1".as_slice(), b"12", b"123", b"1234"] {
             let mut decoder = read::DeflateDecoder::new(input);
-            assert!(std::io::copy(&mut decoder, &mut std::io::sink()).is_err());
+            let error = std::io::copy(&mut decoder, &mut std::io::sink()).unwrap_err();
+            assert_eq!(error.kind(), std::io::ErrorKind::UnexpectedEof);
+            assert_eq!(error.to_string(), "incomplete deflate stream");
         }
     }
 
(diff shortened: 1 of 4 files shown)


Byron:
Thanks a lot!

This looks good to me, and I could validate that the test will fail if the fix is removed.

Byron (approved):

Byron:
Let me CC @jongiddy for a second pair of eyes.

The question could also be if this change introduces downstream breakage, and I'd think it won't except for point out errors that were previously overlooked as they didn't surface.

For that reason, I think merging this is fine, with the option to forward-fix before the next release, which I have no plans for.

PR #557, opened 2026-07-28
feat: imrpove documentation around `ZlibEncoder` trait impls (#554)

### LINKED ISSUE #554: confusing passthrough Read Write implementations
## context

I recently started using `flate2::write::ZlibDecoder` for the first time. While looking at the docs, I wasn't sure how to use it correctly. The doc example shows writing encoded bytes into the decoder and reading decoded bytes through `finish`. This wasn't right for my use case because I was decoding from a continuous stream where the decoder needs to retain state. The correct way of doing this is:

```rust
fn decode_frame(
    decoder: &mut flate2::write::ZlibDecoder<Vec<u8>>,
    frame: &[u8],
) -> std::io::Result<Vec<u8>> {
    decoder.write_all(frame)?;
    decoder.flush()?;
    Ok(std::mem::take(decoder.get_mut()))
}
```

However, I misunderstood `get` and `get_mut` and did not realize that this is the correct approach. I saw in the docs `impl<W: Read + Write> Read for ZlibDecoder<W>`. I thought my `W` needed to implement Read so that I could read decoded bytes through this impl. Resulting in this code:

```rust
fn decode_frame(
    decoder: &mut flate2::write::ZlibDecoder<VecDeque<u8>>,
    frame: &[u8],
) -> std::io::Result<Vec<u8>> {
    decoder.write_all(frame)?;
    decoder.flush()?;
    let mut out = Vec::new();
    decoder.read_to_end(&mut out).unwrap();
}
```

Note that I switched W from Vec to VecDeque. This code actually works! But I think that is a coincidence and unintended. I don't fully understand why it works.

## conclusion

The `impl<W: Read + Write> Read for ZlibDecoder<W>` is undocumented. It was added for a Tokio integration in 2017 related to AsyncRead and AsyncWrite in  39812e6, 9abf34f. It became obsolete in 2022 in 4e5e0c5. I think the impl has served no purpose since then. I'm not confident in my understanding of the crate internals so this part might be wrong.

This isn't limited to `write::ZlibDecoder`. Affects most (all?) encoders and decoders, not just the write ones. 

I would like to make this less confusing. The documentation should explain that this impl is obsolete. Long term the impl should be removed but that is a breaking change.

Byron:
Fixes #554

### DIFF
--- a/src/zlib/write.rs
+++ b/src/zlib/write.rs
@@ -180,10 +180,16 @@ impl<W: Read + Write> Read for ZlibEncoder<W> {
 /// This structure implements a [`Write`] and will emit a stream of decompressed
 /// data when fed a stream of compressed data.
 ///
+/// If the underlying writer also implements [`Read`], this type forwards reads
+/// directly to it; reading from the decoder does not perform decompression. To
+/// retrieve output while retaining the decoder's state, flush the decoder and
+/// access the writer with [`get_mut`](ZlibDecoder::get_mut).
+///
 /// After decoding a single member of the ZLIB data this writer will return the number of bytes up
 /// to the end of the ZLIB member and subsequent writes will return Ok(0) allowing the caller to
 /// handle any data following the ZLIB member.
 ///
+/// [`Read`]: https://doc.rust-lang.org/std/io/trait.Read.html
 /// [`Write`]: https://doc.rust-lang.org/std/io/trait.Write.html
 ///
 /// # Examples


e00E:
So you're updating the docs for `flate2::write::ZlibDecoder`. That's the example from my issue but the same thing applies to a lot of types. The dual trait thing applies to all encoders and decoders, not just the write ones. It's probably only confusing for the write ones though.

Byron:
Right, as pointed out, this is documented on library level and we can chose to put this note to all related types if more maintainers think it's a good idea duplicate documentation like that.

Also reference: https://github.com/rust-lang/flate2-rs/issues/554#issuecomment-5102434539

PR #558, opened 2026-07-28
Update `get_mut()` documentation to make clear how 'inner` can be used

### LINKED ISSUE #555: Confusing warning on
Several types in this crate have a `get_mut` method like `flate2::write::ZlibDecoder::get_mut` with a doc comment saying:

> Note that mutating the output/input state of the stream may corrupt this object, so care must be taken when using this method.

I do not understand this comment. What care must be taken? What can and can't I do? The comment should explain this. I can't confidently use this method correctly without more details.

For context, the way I think you are supposed to use this method is like this, when you have a decoder that you need to reuse between decodes.

```rust
fn decode_frame(
    decoder: &mut flate2::write::ZlibDecoder<Vec<u8>>,
    frame: &[u8],
) -> std::io::Result<Vec<u8>> {
    decoder.write_all(frame)?;
    decoder.flush()?;
    Ok(std::mem::take(decoder.get_mut()))
}
```

Byron:
This should help steering towards `reset()` in places where `flush() + swap(get_mut, x)` might otherwise be used.

It also mentions how to use `get_mut()` for streaming.

Fixes #555

### DIFF
--- a/src/deflate/write.rs
+++ b/src/deflate/write.rs
@@ -52,11 +52,13 @@ impl<W: Write> DeflateEncoder<W> {
     ///
     /// The underlying writer may be mutated or replaced as long as this
     /// preserves the bytes and ordering of the logical output stream.
+    /// Concatenate output from each writer to reconstruct the complete stream.
     ///
-    /// For streaming output, call [`flush`](Write::flush) before replacing the
-    /// writer, such as with [`std::mem::take`], to retrieve all output produced
-    /// so far. Concatenate output from each writer to reconstruct the complete
-    /// stream.
+    /// Replacing the writer does not require [`flush`](Write::flush). Call it
+    /// first when all input accepted so far must be decodable without output
+    /// from later writes. This inserts a sync-flush point and changes the output
+    /// bitstream. This is useful before applying [`std::mem::take`] to
+    /// [`get_mut`](Self::get_mut) when forwarding the stream incrementally.
     ///
     /// To start a new stream, use [`reset`](Self::reset); replacing the writer
     /// does not reset this encoder.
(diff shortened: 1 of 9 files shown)


src/deflate/write.rs
```diff
@@ -50,8 +50,16 @@ impl<W: Write> DeflateEncoder<W> {
...
     ///
-    /// Note that mutating the output/input state of the stream may corrupt this
-    /// object, so care must be taken when using this method.
+    /// The underlying writer may be mutated or replaced as long as this
+    /// preserves the bytes and ordering of the logical output stream.
+    ///
+    /// For streaming output, call [`flush`](Write::flush) before replacing the
+    /// writer, such as with [`std::mem::take`], to retrieve all output produced
+    /// so far. Concatenate output from each writer to reconstruct the complete
+    /// stream.
```
jongiddy: This perhaps over-emphasizes the need for flushing. You can replace a `Write` implementation at any time without affecting the state of the Encoder/Decoder. To obtain the full output you must (at least logically) concatenate the output of all writers in order, possibly requiring a flush on each of them. However, it is not required to flush the Encoder/Decoder before each replacement.

Maybe something more like:
```suggestion
    /// The underlying writer may be mutated or replaced as long as this
    /// preserves the bytes and ordering of the logical output stream. 
    /// Concatenate output from each writer to reconstruct the complete
    /// stream.
    ///
    /// To ensure output is retrieved as soon as it is available, call [`flush`](Write::flush) 
    /// before replacing the writer.
```

jongiddy (approved):

src/deflate/write.rs
fintelia: The part that neither of these descriptions really express is that calling flush _changes_ the output bitstream by inserting a sync-flush. That's necessary to ensure that all input bytes passed in prior to the flush can be decoded without needing further bytes of output produced by any calls following the flush.

Byron: Thanks @jongiddy and @fintelia! Particularly the "flush has side-effects" note is something the caller should definitely be aware of.

What about this?

##### Encoders

```rust
    /// Acquires a mutable reference to the underlying writer.
    ///
    /// The underlying writer may be mutated or replaced as long as this
    /// preserves the bytes and ordering of the logical output stream.
    /// Concatenate output from each writer to reconstruct the complete stream.
    ///
    /// Replacing the writer does not require [`flush`](Write::flush). Call it
    /// first when all input accepted so far must be decodable without output
    /// from later writes. This inserts a sync-flush point and changes the output
    /// bitstream. This is useful before applying [`std::mem::take`] to
    /// [`get_mut`](Self::get_mut) when forwarding the stream incrementally.
    ///
    /// To start a new stream, use [`reset`](Self::reset); replacing the writer
    /// does not reset this encoder.
```

##### Decoders

```rust
    /// Acquires a mutable reference to the underlying writer.
    ///
    /// The underlying writer may be mutated or replaced as long as this
    /// preserves the bytes and ordering of the logical output stream.
    /// Concatenate output from each writer to reconstruct the complete stream.
    ///
    /// Replacing the writer does not require [`flush`](Write::flush). Call it
    /// first to write all decompressed output currently available to the
    /// current writer. This is useful before applying [`std::mem::take`] to
    /// [`get_mut`](Self::get_mut) when forwarding output incrementally.
    ///
    /// To start a new stream, call [`finish`](Self::finish) and create a new
    /// decoder; replacing the writer does not reset it.
```

I have also updated the PR to reflect the change everywhere, with both forms.

PR #559, opened 2026-08-04
Add `no_std` support using nightly `alloc_io`

bushrat011899:
# Description

With [rust-lang/rust#154046](https://github.com/rust-lang/rust/issues/154046) implemented, `alloc::io` is available on nightly. With this, `flate2` could add meaningful `no_std` support. As it is currently only available on nightly, I've gated `no_std` support behind an unstable rustflag, `flate2_unstable_nightly_alloc_io`. With this, `flate2` becomes compatible with targets such as `wasm32v1-none`.

If/when `alloc_io` is stabilised as a feature, `flate2_unstable_nightly_alloc_io` could be replaced with a stable feature flag alternative (e.g., `no_std`) which would simply increase MSRV. Alternatively, a new version of `flate2` with an increased MSRV would allow `no_std` support to be unconditional.

---

## Notes

* No AI tooling of any kind was used during the creation of this PR.


bushrat011899:
Confirmed CI failure is unrelated to this PR. Opened #561

Byron (changes requested):
Thanks!

I took a quick look everything looks pretty straightforward and non-invasive.
CI should tests this new feature toggle though. Maybe there is other tests/validation you can think of as well, like what it means to work in `wasm32v1-none`?

<details><summary>Local Codex Review</summary>
<p>

What follows is unvalidated, you'd have to double-check.

----

The patch introduces a breaking public API change for standard users and disables runtime hardware acceleration in two backends. These regressions should be addressed before merging the no_std support.

Full review comments:

- [P1] Restore the SystemTime return type on std builds — /Users/byron/dev/github.com/rust-lang/flate2-rs.no_std/src/gz/mod.rs:82-87
  For every normal `std` build, this public method now returns a duration since the epoch instead of a `SystemTime`, so downstream code expecting `SystemTime` or calling its methods no longer compiles. Preserve the existing `SystemTime` API when `std` is available and expose a duration-based alternative only for `no_std` builds.

- [P2] Preserve zlib-rs runtime CPU detection — /Users/byron/dev/github.com/rust-lang/flate2-rs.no_std/Cargo.toml:25-25
  When a standard build selects `zlib-rs` on a baseline target, replacing `std` with `rust-allocator` disables zlib-rs's `cfg(feature = "std")` runtime SIMD detection. This does not enable `no_std` either: in zlib-rs 0.6, `rust-allocator` itself causes the crate to link `std`. Keep `std` for normal builds; a genuine `no_std` path needs a non-std allocator integration.

- [P2] Preserve CRC runtime dispatch for std builds — /Users/byron/dev/github.com/rust-lang/flate2-rs.no_std/Cargo.toml:29-29
  For normal builds targeting baseline x86_64 or an ARM target without CRC in its baseline features, disabling `crc32fast`'s `std` feature replaces runtime hardware detection with compile-time `target_feature` checks. Published binaries will therefore use the software CRC path even on capable CPUs, regressing gzip throughput; keep `crc32fast/std` enabled for the default `std` configuration and disable it only for `no_std`.

</p>
</details>

bushrat011899:
Thanks for the quick review! I've made a few changes based on your and Codex's feedback:

1. I've added a (default) `std` feature which pulls double-duty in enabling runtime CPU feature detection in `zlib-rs` and `crc32fast` and to control the availability of `GzHeader::mtime_as_datetime`. The `Duration` based function I added previously is now a new function named `GzHeader::mtime_as_duration`. It's not strictly necessary, but I think it's a nice-to-have with very low risk. Happy to remove if it's a problem!
2.  Added a new `no_std` CI job which checks for (nightly) compatibility on `wasm32v1-none`, `x86_64-unknown-none`, and `thumbv4t-none-eabi`. These three targets give a really good breadth of `no_std` use cases (maximum compatibility web, freestanding desktop, and embedded respectively). All three targets have no `libstd` artifact available, so if a crate can build (or even just `check`) on such a target successfully, that proves the crate and all active dependencies are `no_std` compatible.

Byron:
This looks even better, thanks!

Now I wonder: What's the value proposition here? That the crate can build, or that it can be used to do work? Since it's probably the latter, I wonder why CI isn't running tests?
And once it does, I think it should run the test-suite against multiple backends similar to what's [already done](https://github.com/rust-lang/flate2-rs/blob/00864eec25b103550b63d9be9c96361da0b12e55/.github/workflows/main.yml#L37-L52).
Thanks again.

## Codex Review

- [P1] Preserve the datetime method in std-backed builds — /Users/byron/dev/github.com/rust-lang/flate2-rs.no_std/src/gz/mod.rs:82-83
  For existing stable configurations such as `default-features = false, features = ["zlib"]`, `flate2_unstable_nightly_alloc_io` is unset and the crate still links and uses `std` via `src/lib.rs:130-140`, but this new feature gate removes the previously available `mtime_as_datetime` method and breaks downstream compilation. Gate it with the same `any(feature = "std", not(flate2_unstable_nightly_alloc_io))` condition so only actual no-std builds lose the `SystemTime` API.

bushrat011899:
> This looks even better, thanks!
> 
> Now I wonder: What's the value proposition here? That the crate can build, or that it can be used to do work? Since it's probably the latter, I wonder why CI isn't running tests? And once it does, I think it should run the test-suite against multiple backends similar to what's [already done](https://github.com/rust-lang/flate2-rs/blob/00864eec25b103550b63d9be9c96361da0b12e55/.github/workflows/main.yml#L37-L52). Thanks again.

Purely an oversight on my part! `no_std` targets (e.g., `wasm32v1-none`) don't have the `test` crate, so you can't run `cargo test` on them. So I've changed the CI to test the `no_std` configuration on an existing platform. So the `no_std` job is purely to check that the crate can compile, and then the existing test job actually ensures the `no_std` configuration works as expected.

In response to the Codex review, I've adjusted the gating (again). Instead of an `std` feature, I've renamed it `runtime_detection`, since (today), disabling the feature-formerly-known-as `std` wouldn't actually make the crate `no_std` compatible, which is confusing. I've then changed the gate on `mtime_as_datetime` to `not(flate2_unstable_nightly_alloc_io)`. Since `RUSTFLAGS` are exempt from the additive nature of features, it's not a breaking change to config out that function.

I've also split this PR into two commits, one which just makes stable changes (import from `core` and `alloc` instead of `std`, etc.). So the unstable nightly stuff is now more focussed in a single commit, which should make it easier to revert (if needed) in the future.

src/crc.rs
```diff
@@ -319,7 +319,8 @@ mod tests {
 
         ca.combine(&cb);
 
-        dbg!(&ca);
+        #[cfg(not(flate2_unstable_nightly_alloc_io))]
+        std::dbg!(&ca);
```
Byron: It looks like this stray `dbg!` shouldn't be in the code to begin with. Probably worth removing.

Cargo.toml
```diff
@@ -22,11 +22,11 @@ exclude = [".*"]
 libz-sys = { version = "1.1.20", optional = true, default-features = false }
 libz-ng-sys = { version = "1.1.16", optional = true }
 # this matches the default features, but we don't want to depend on the default features staying the same
```
jongiddy: This comment doesn't appear to be accurate even before this change, but is definitely not accurate now. It can be removed.

src/crc.rs
bushrat011899: Yeah I thought so but erred on the side of caution, happy to delete!

Cargo.toml
bushrat011899: Good catch, deleted!

bushrat011899:
> I feel like anything we do with features is going to be a breaking change in the light of people using `default-features = false`, at least if performance is a part of it.

According to the [Cargo Book](https://doc.rust-lang.org/cargo/reference/semver.html#cargo-change-dep-feature), removing features from a dependency is only considered a minor change, so long as the public API does not also change. "Technically" not a breaking change probably isn't a great selling point though. I think the impact should be relatively minor, especially since the mitigation is as simple as adding a feature flag, which end users can do even if `flate2` is only used transitively.

> With that said, I'd love it if the commit messages would be more descriptive, also to explain more what the commit enables, its motivation, if you will.

Perfectly reasonable, especially now that I think the final form of this PR appears to have taken shape. I've split apart the commits to be more focussed and added details to better motivate why a change is made. Each is atomic, so functionality like `git bisect` should be unaffected.

.github/workflows/main.yml
```diff
@@ -93,6 +99,32 @@ jobs:
...
+    - name: Install Target
+      run: rustup target add ${{ matrix.target }}
+      if: matrix.target != 'thumbv4t-none-eabi'
+    - name: Install rust-src
+      run: rustup component add rust-src
+      if: matrix.target == 'thumbv4t-none-eabi'
+      # Tests require access to the `test` crate, which requires `std`. The goal
+      # of this job is to simply confirm compilation on `no_std` targets succeeeds.
+      # Testing the `no_std` confirmation is done above.
+    - run: cargo build ${{ matrix.flags }} --target ${{ matrix.target }} --no-default-features --features rust_backend
```
Byron, line 126: This feels a little bit like giving up too early.
Can't we create a little WASM program that compresses and decompresses a stream, to have a clear life-sign?
Maybe examples/ can already be used for this, or maybe there would have to be a special WASM example to make this work?

bushrat011899: Ok, it's a bit of work, but I've added a runnable `no_std` example which is checked for correct outputs in CI. The simplest environment I could think to get setup to actually execute `no_std` code is UEFI, so I've made a little example project in the examples folder which can build an run a UEFI application within QEMU. The actual code it runs is based off of one of the existing examples, and I've tried to separate out the `no_std` special ceremony from the "normal" Rust code to make maintenance as simple as possible.

In CI, I test that example on QEMU with the `x86_64-unknown-uefi` target, and I also run it on `x86_64-unknown-linux-gnu` (the standard desktop Linux target) to ensure it produces the same results in both cases.

Byron: Oh wow, sorry for that, I thought it was more straightforward. When I read QEMU, I read complications 😅.

I think what I had in mind was merely to use some wasm runtime that runs a wasm executable with a compress/decompress roundtrip.

But now that you went through the trouble, it seems like a very nice thing to have the QEMU version, as stand-in for embedded probably.

